PR #41, opened 2018-11-06
decode.c fixes

DenisPryt:
If `end - dat->byte == 1` in `dwg_decode_eed`, `obj->num_eed` will be increased, but realloc don't invokes.

According to specification section_size can't be greater then 2032. But in fact, there are files with a section size of 2035.


rurban:
Nice. I've always suspect a wrong realloc there, but couldn't find it.

PR #42, opened 2018-11-13
Bug fixes

DenisPryt:
Fix some warnings and crashes: bracers in defines (out_geojson.c), pass a correct parameters to fwrite, Bit_Chain initialization, read section name according to name_length, free section name, check num_pages for zero (avoiding malloc(0)), fix posible memory leeks in dwg_api.c and reedsolomon.c

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -3015,7 +3015,7 @@ get_base_value_type(short gc)
               if (gc <= 469) return VT_REAL;
               if (gc <= 479) return VT_STRING;
               if (gc <= 998) return VT_INVALID;
-              /*if (gc == 999)*/ return VT_STRING;
+              if (gc == 999) return VT_STRING; /* dead if */
             }
         }
       else // <440
@@ -3026,7 +3026,7 @@ get_base_value_type(short gc)
               if (gc <= 409) return VT_INT16;
               if (gc <= 419) return VT_STRING;
               if (gc <= 429) return VT_INT32;
-              /*if (gc <= 439)*/ return VT_STRING;
+              if (gc <= 439) return VT_STRING; /* dead if */
             }
           else            // 330-389
             {
@@ -3034,7 +3034,7 @@ get_base_value_type(short gc)
               if (gc <= 319) return VT_BINARY;
               if (gc <= 329) return VT_HANDLE;
               if (gc <= 369) return VT_OBJECTID;
-              /*if (gc <= 389)*/ return VT_INT16;
+              if (gc <= 389) return VT_INT16; /* dead if */
             }
         }
     }
@@ -3046,7 +3046,7 @@ get_base_value_type(short gc)
           if (gc <= 269) return VT_INVALID;
           if (gc <= 279) return VT_INT16;
           if (gc <= 289) return VT_INT8;
-          /*if (gc <= 299)*/ return VT_BOOL;
+          if (gc <= 299) return VT_BOOL; /* dead if */
         }
       else               // 105-209
         {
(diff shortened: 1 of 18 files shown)


src/decode.c
```diff
@@ -146,7 +146,7 @@ dwg_decode(Bit_Chain * dat, Dwg_Data * dwg)
   memset(&dwg->header_vars, 0, sizeof(Dwg_Header_Variables));
   memset(&dwg->r2004_header.file_ID_string[0], 0, sizeof(dwg->r2004_header));
   memset(&dwg->auxheader.aux_intro[0], 0, sizeof(dwg->auxheader));
-  memset(&dwg->second_header.size, 0, sizeof(dwg->second_header));
+  memset(&dwg->second_header, 0, sizeof(dwg->second_header));
```
rurban: I remember vaguely there was a reason to only clear the size. Need to check

src/decode.c
```diff
@@ -3016,7 +3015,7 @@ get_base_value_type(short gc)
               if (gc <= 469) return VT_REAL;
               if (gc <= 479) return VT_STRING;
               if (gc <= 998) return VT_INVALID;
-              if (gc == 999) return VT_STRING;
+              /*if (gc == 999)*/ return VT_STRING;
```
rurban: why?

src/decode.c
```diff
@@ -3027,15 +3026,15 @@ get_base_value_type(short gc)
               if (gc <= 409) return VT_INT16;
               if (gc <= 419) return VT_STRING;
               if (gc <= 429) return VT_INT32;
-              if (gc <= 439) return VT_STRING;
+              /*if (gc <= 439)*/ return VT_STRING;
```
rurban: why? doesn't look right

src/decode_r2007.c
```diff
@@ -157,7 +157,7 @@ static void copy_bytes(char *dst, uint32_t length, uint32_t offset);
 static uint32_t read_literal_length(unsigned char **src, unsigned char opcode);
 static void copy_compressed_bytes(char *restrict dst, char *restrict src, int length);
 static void  bfr_read(void *restrict dst, char **restrict src, size_t size);
-static DWGCHAR* bfr_read_string(char **src);
+static DWGCHAR* bfr_read_string(char **src, int64_t stringSize);
```
rurban: good. it's static so it needs no version bump. 0.6.3 is fine

src/logging.h
```diff
@@ -91,7 +91,7 @@
   if (DWG_LOGLEVEL >= DWG_LOGLEVEL_##level && wstr) { \
     BITCODE_TU ws = wstr;                             \
     uint16_t _c;                                      \
-    while ((_c = *ws++)) {                            \
+    while (ws && (_c = *ws++)) {                      \
```
rurban: Not needed. We already checked wstr in line 91

DenisPryt: Yor are right. I'll remove this.

src/decode.c
DenisPryt: size is the first field of the structure and sizeof all header. 
I think `&dwg->second_header.size` and `&dwg->second_header are equals`.
Maybe correct code is `memset(&dwg->second_header.size, 0, sizeof(dwg->second_header.size));`?

src/decode.c
DenisPryt: This not a mistake, just unnecessary `if`. This comparison is always true

src/decode.c
DenisPryt: Not a mistake. Just this comparison is always true

DenisPryt:
I'm waiting for recommendations on `get_base_value_type` and `decode.c`

src/decode.c
rurban: Oh, good call. I didn't see that.

src/decode.c
rurban: I see. But I rather want to keep it in nevertheless then, please.

src/decode.c
rurban: Your patch is fine then

rurban:
Merged with e254c3b2d91370c32dd6e1b50b109c6aee6437d4

PR #45, opened 2018-11-23
Corrupted eed decode fix

DenisPryt:
`dwg_decode_eed` function, has a `while ((size = bit_read_BS(dat)))` loop and variable `end`
`end` is ignore last `size = bit_read_BS(dat)`, which must be zero.
If `eed` section is present and corrupted, happens seek into `end`, and `return` inside a while loop.
As result, we miss the last `size = bit_read_BS(dat)` in `while` condition.


rurban:
Thanks. I'll fix up the warnings and whitespace errors by myself.

Without the move it's still a tiny change, but can you please sign the legal GNU papers?
Would you be willing to assign the copyright to the Free Software Foundation, so that we could install it?
See https://www.gnu.org/prep/maintain/html_node/Copyright-Papers.html#Copyright-Papers
and `CONTRIBUTING`

DenisPryt:
Yes of course. But I think, i make a mistake with `size` in my code (I calculate it in my function, and it a mistake)

rurban:
I'll push my fixes to `smoke/eed`. Please continue fixing it from there

DenisPryt:
Sure

PR #48, opened 2018-11-26
Fix error with eed size

DenisPryt:
Fix my bug with wrong eed section size


rurban:
Thanks, moving this around looks good.

I'm just wondering what happened to the Appveyor check. I might have broken it.

PR #49, opened 2018-11-26
Fix multileader bug

DenisPryt:
Fix `DWG_ENTITY(MULTILEADER)` spec bug. Vertexes should be array.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,165 @@
 Full history from the git log
 -----------------------------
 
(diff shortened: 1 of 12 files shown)


rurban:
Ouch, a good one! But please without the merge commit

rurban:
See my branch `smoke/multileader`

DenisPryt:
I merged your smoke/multileader branch with my changes, but your branch has a lot of changes, except `FIELD_3DPOINT_VECTOR ` and `line. numpts` naming. It's ok?

PR #54, opened 2018-11-27
Fix field value decode

DenisPryt:
I am not sure, that this right fixes. 
Page 241 in [spec](https://www.opendesign.com/files/guestdownloads/OpenDesign_Specification_for_.dwg_files.pdf) comment to `512 – General`


rurban:
Will be merged together with #58

rurban:
Merged into master. Thanks

PR #58, opened 2018-11-28
Fix memory leaks

DenisPryt:
When eed decoding error occurs, the eed is freed incorrectly.
DWG_TYPE_FREED collision.
Any `return` in dwg.spec leads to memory leaks, because some frees was in OBJECT_END ENTITY_END macroses.


DenisPryt:
I can't understand what's wrong

rurban:
I don't like the _private thing.
If we need to jump to DWG_OBJECT_END from early exit, we can use a goto object_end_#token; instead of a return.

DenisPryt:
I realized that `return` inside a `REPEAT` section leads to memory leak, becouse `free` array occurs in `END_REPEAT` macros.

DenisPryt:
What a reason to remove a broken `Dwg_Object_Ref*` from `object_ref`? It leads to `Dwg_Object_Ref*` leaks.
There are a 4 places in `decode.c` with this type of leaks.
```
switch (ref->handleref.code)
{
   //...
   default:
      dwg->object_ref[dwg->num_object_refs-1] = NULL;
      dwg->num_object_refs--;
      ref->absolute_ref = 0;
      ref->obj = NULL;
}
```
and
```
// If the handle size is 0 and not a relative handle, it is probably
// a null handle.
// It shouldn't be placed in the object ref vector.
if (ref->handleref.size || (obj && ref->handleref.code > 5))
{
   //...
   dwg->object_ref[dwg->num_object_refs++] = ref;
}
else if (!ref->handleref.value)
{
  ref->obj = NULL;
  ref->absolute_ref = 0;
  return ref;
}
```

Maybe true way is free broken refs and return `NULL`? Or make a second array with broken refs? Or just add them to `object_ref` in spite of comments?

rurban:
Ok, I've warmed up to the _private functions. The leaks are all gone.
But you need to declare them before.
And the new reactors overflow check is wrong.

src/dwg.spec
```diff
@@ -3485,11 +3498,15 @@ DWG_ENTITY(HATCH)
...
+                      #ifdef IS_FREE
+                        END_REPEAT(paths[rcount1].segs);
+                        END_REPEAT(paths);
+                      #endif
                       return DWG_ERR_VALUEOUTOFBOUNDS;
                 }
             }
           SET_PARENT(paths[rcount1].segs, &_obj->paths[rcount1])
-          END_REPEAT(paths[rcount1].segs);
+		  END_REPEAT(paths[rcount1].segs);
```
rurban: formatting

src/dwg.spec
```diff
@@ -4086,6 +4105,9 @@ DWG_OBJECT(FIELD)
   //DEBUG_HERE_OBJ
   FIELD_T (evaluation_error_msg, 300);
   Table_Value(value)
+  if (error & ~DWG_ERR_INVALIDTYPE)
```
rurban: logical error (and below):
why are you checking for all errors but INVALIDTYPE and return INVALIDTYPE then?
I think you need to return error instead.

src/dwg.spec
```diff
@@ -1488,7 +1488,6 @@ DWG_ENTITY(SPLINE)
 
   if (FIELD_VALUE(scenario) & 1) {
     FIELD_VECTOR(knots, BD, num_knots, 40)
-    END_REPEAT(knots);
```
rurban: good

src/decode_r2007.c
```diff
@@ -609,7 +609,7 @@ read_system_page(Bit_Chain* dat, int64_t size_comp, int64_t size_uncomp,
                 (long)page_size, dat->size - dat->byte);
       return NULL;
     }
-  data = (BITCODE_RC*)calloc(size_uncomp, page_size);
+  data = (BITCODE_RC*)calloc(size_uncomp + page_size, 1);
```
rurban: shouldn't this be *?

src/free.c
```diff
@@ -291,27 +291,27 @@ dwg_free_xdata(Dwg_Object_XRECORD *obj, int size)
   obj->xdata = NULL;
 }
 
-static void
+void
```
rurban: you need to EXPORT it then

src/dec_macros.h
```diff
@@ -751,20 +767,36 @@ EXPORT int dwg_add_ ## token (Dwg_Object *obj) \
   obj->tio.object->objid = obj->index; /* obj ptr itself might move */ \
   return 0; \
 } \
+\
```
rurban: this needs the _private decl here:
```
static int dwg_decode_ ## token ## _private (Bit_Chain *dat, Bit_Chain *str_dat, \
                                             Dwg_Object *restrict obj); \
\
```

src/dwg.spec
```diff
@@ -4094,6 +4116,11 @@ DWG_OBJECT(FIELD)
     {
       FIELD_T (childval[rcount1].key, 6);
       Table_Value(childval[rcount1].value)
+      if (error & ~DWG_ERR_INVALIDTYPE)
+        {
+          END_REPEAT(childval)
```
rurban: whitespace at the eol

src/dwg.spec
```diff
@@ -1716,17 +1717,38 @@ static int encode_3dsolid(Bit_Chain* dat, Bit_Chain* hdl_dat, Dwg_Object* obj,
 static void free_3dsolid(Dwg_Object* obj, Dwg_Entity_3DSOLID* _obj)
```
rurban: this needs to be `static int free_3dsolid` then

src/decode.c
```diff
@@ -2779,6 +2779,27 @@ dwg_decode_object(Bit_Chain* dat, Bit_Chain* hdl_dat, Bit_Chain* str_dat,
   return error;
 }
 
+static int
+dwg_decode_add_object_ref(Dwg_Data *restrict dwg, Dwg_Object_Ref *ref)
+{
+  const Dwg_Object_Ref **object_ref_old = dwg->object_ref;
```
rurban: type error: const Dwg_Object_Ref *object_ref_old

src/dwg.spec
rurban: and free_3dsolid needs to be moved to src/free.c

src/free.c
rurban: this broke decode_test:
/usr/bin/ld: decode_test.o: in function `dwg_decode_eed':
/home/rurban/Software/libredwg/test/testcases/./../../src/decode.c:2428: undefined reference to `dwg_free_eed'
/usr/bin/ld: /home/rurban/Software/libredwg/test/testcases/./../../src/decode.c:2413: undefined reference to `dwg_free_eed'
clang: error: linker command failed with exit code 1 (use -v to see invocation)

src/decode_r2007.c
DenisPryt: It look like 2 arrays. `data` and `rsdata `
```
rsdata = &data[size_uncomp];
bit_read_fixed(dat, rsdata, page_size);
//...
decompress_r2007(data, size_uncomp, pedata, size_comp);
```

src/dwg.spec
DenisPryt: Ooops, I wanted to write `if (error & DWG_ERR_INVALIDTYPE)` (check only DWG_ERR_INVALIDTYPE error)
Ok, it will be `if (error) return error;`

rurban:
Merged into master with my fixups, Thanks

DenisPryt:
I did not find a `denis/memory_leaks` branch. What wrong with _VECTOR_CHECHOUT in REACTORS?

rurban:
Already deleted and merged into master. See a319f6318a9e24b8ba96d2555e23c4ba2c98584e

PR #65, opened 2018-12-07
Fixes REPEAT_CHKCOUNT, section_handles and obj_string_stream

DenisPryt:
Add a minimal bit sizes to difficult tupes, used with `REPEAT` macros.
Fix `read_section_handles`: `bit_read_MC` can consist of 5 bytes. There is on limits in opendesign.com, but I have files with 5'th byte offsets `10000000`. => #70 
Add crc check to read section handles.
Minor fixes `free` -> `FREE_IF`


src/dwg.spec
```diff
@@ -2122,22 +2116,9 @@ DWG_ENTITY(MLINE)
...
           FIELD_BS (verts[rcount1].lines[rcount2].num_segparms, 74);
-          REPEAT3(verts[rcount1].lines[rcount2].num_segparms,
-                  verts[rcount1].lines[rcount2].segparms,
-                  BITCODE_BD)
-            {
-              FIELD_BD (verts[rcount1].lines[rcount2].segparms[rcount3], 41);
-            }
-          END_REPEAT(verts[rcount1].lines[rcount2].segparms);
-
+          FIELD_VECTOR(verts[rcount1].lines[rcount2].segparms, BD, verts[rcount1].lines[rcount2].num_segparms, 41);
```
rurban: Good

src/dwg.spec
```diff
@@ -1489,21 +1490,14 @@ DWG_ENTITY(SPLINE)
...
-      }
-    END_REPEAT(ctrl_pts);
+	SPLINE_CONTROL_POINT_VECTOR(ctrl_pts, num_ctrl_pts, weighted)
   }
   if (FIELD_VALUE(scenario) & 2) {
     REPEAT(num_fit_pts, fit_pts, Dwg_SPLINE_point)
       {
         FIELD_3BD (fit_pts[rcount1], 11);
       }
+	SET_PARENT_OBJ(fit_pts);
```
rurban: good

src/dwg.spec
```diff
@@ -2122,22 +2116,9 @@ DWG_ENTITY(MLINE)
...
+          FIELD_VECTOR(verts[rcount1].lines[rcount2].segparms, BD, verts[rcount1].lines[rcount2].num_segparms, 41);
           FIELD_BS (verts[rcount1].lines[rcount2].num_areafillparms, 75);
-          REPEAT3(verts[rcount1].lines[rcount2].num_areafillparms,
-                  verts[rcount1].lines[rcount2].areafillparms,
-                  BITCODE_BD)
-            {
-              FIELD_BD (verts[rcount1].lines[rcount2].areafillparms[rcount3], 42);
-            }
-          END_REPEAT(verts[rcount1].lines[rcount2].areafillparms);
+          FIELD_VECTOR(verts[rcount1].lines[rcount2].areafillparms, BD, verts[rcount1].lines[rcount2].num_areafillparms, 42);
```
rurban: Good

src/dwg.spec
```diff
@@ -3554,11 +3542,7 @@ DWG_ENTITY(HATCH)
           FIELD_2BD_1 (deflines[rcount1].pt0, 43);
           FIELD_2BD_1 (deflines[rcount1].offset, 45);
           FIELD_BS (deflines[rcount1].num_dashes, 79);
-          REPEAT2 (deflines[rcount1].num_dashes, deflines[rcount1].dashes, BITCODE_BD)
-            {
-              FIELD_BD (deflines[rcount1].dashes[rcount2], 49);
-            }
-          END_REPEAT(deflines[rcount1].dashes);
+          FIELD_VECTOR(deflines[rcount1].dashes, BD, deflines[rcount1].num_dashes, 49);
```
rurban: Good

src/dwg.spec
```diff
@@ -5910,7 +5894,8 @@ DWG_ENTITY(HELIX)
   FIELD_BL (scenario, 0);
   UNTIL(R_2013) {
     if (FIELD_VALUE(scenario) != 1 && FIELD_VALUE(scenario) != 2)
-      fprintf(stderr, "Error: unknown scenario %d", FIELD_VALUE (scenario));
+      LOG_ERROR("Error: unknown scenario %d", (int)FIELD_VALUE(scenario));
+      return DWG_ERR_INVALIDTYPE;
```
rurban: Good

src/dwg.spec
```diff
@@ -5958,21 +5943,14 @@ DWG_ENTITY(HELIX)
...
-      }
-    END_REPEAT(ctrl_pts);
+    SPLINE_CONTROL_POINT_VECTOR(ctrl_pts, num_ctrl_pts, weighted)
   }
   if (FIELD_VALUE(scenario) & 2) {
     REPEAT(num_fit_pts, fit_pts, Dwg_SPLINE_point)
       {
         FIELD_3BD (fit_pts[rcount1], 11);
       }
+	SET_PARENT_OBJ(fit_pts);
```
rurban: Good

src/dwg.spec
```diff
@@ -1441,7 +1441,8 @@ DWG_ENTITY(SPLINE)
   FIELD_BL (scenario, 0);
   UNTIL(R_2013) {
     if (FIELD_VALUE(scenario) != 1 && FIELD_VALUE(scenario) != 2)
-      fprintf(stderr, "Error: unknown scenario %d", FIELD_VALUE (scenario));
+      LOG_ERROR("Error: unknown scenario %d", (int)FIELD_VALUE(scenario));
```
rurban: Good

src/dwg.spec
```diff
@@ -5958,21 +5943,14 @@ DWG_ENTITY(HELIX)
...
-    REPEAT(num_ctrl_pts, ctrl_pts, Dwg_SPLINE_control_point)
-      {
-        FIELD_3BD (ctrl_pts[rcount1], 10);
-        if (!FIELD_VALUE(weighted))
-          FIELD_VALUE(ctrl_pts[rcount1].w) = 0; // skipped when encoding
-        else
-          FIELD_BD (ctrl_pts[rcount1].w, 41);
-      }
-    END_REPEAT(ctrl_pts);
+    SPLINE_CONTROL_POINT_VECTOR(ctrl_pts, num_ctrl_pts, weighted)
```
rurban: I'll rather keep the REPEAT logic here. It's only used twice.

src/dwg.spec
```diff
@@ -2482,38 +2463,45 @@ DWG_OBJECT(LTYPE)
     FIELD_RC (alignment, 72);
   }
   FIELD_RC (num_dashes, 73);
-  REPEAT_C(num_dashes, dash, Dwg_LTYPE_dash)
+  PRE(R_13)
```
rurban: Does this fix the LTYPE.dash problem?

DenisPryt:
```
This change don't fix any problem. Just in one situation minimal size is `352`, and in other  -- `138`. More accurate `_VECTOR_CHKCOUNT` and remove `if` from loop.
```

rurban:
Maybe split it up into individual commits: UMC, FREE_IF, obj_string_stream overflow, bit_check_CRC + size

DenisPryt:
Ok, i'll do this

rurban:
Thanks, much better!

src/free.c
```diff
@@ -668,7 +667,7 @@ dwg_free_object(Dwg_Object *obj)
             }
           else // not a class
             {
-              free(obj->tio.unknown);
+              FREE_IF(obj->tio.unknown);
               obj->tio.unknown = NULL;
```
rurban: This line can be removed then also

src/common.c
```diff
@@ -124,6 +124,31 @@ const char *dwg_bits_name[] =
   "3DD",
   "CRC",
   "CRC64",
+  "SPLINE_control_point",
```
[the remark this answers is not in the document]
DenisPryt: Do you want create another one array with `SPLINE_` like structure's minimal sizes and use it in REPEAT macros?
Or any other ideas?

rurban:
I've merged 2 of those commits: 
* free -> FREE_IF migration 
* Fix object-map section decode: offset is BITCODE_UMC with a fixed subject

CRC can probably be cherry-picked also

src/bits.c
```diff
@@ -1147,22 +1147,32 @@ bit_read_CRC(Bit_Chain * dat)
...
+      LOG_WARN("bit_check_CRC out of range");
+      return 0;
+    }
 
-  calculated = bit_calc_CRC(seed, &(dat->chain[start_address]),
-                            dat->byte - start_address);
+  old_byte = dat->byte;
+  old_bit = dat->bit;
 
+  calculated = bit_calc_CRC(seed, &(dat->chain[start_address]), size);
```
rurban: No. This is wrong for unaligned bits. We need to align dat first, as in `if (dat->bit > 0) dat->byte++; dat->bit = 0;`

src/bits.c
```diff
@@ -1147,22 +1147,32 @@ bit_read_CRC(Bit_Chain * dat)
...
-bit_check_CRC(Bit_Chain * dat, long unsigned int start_address,
-              uint16_t seed)
+bit_check_CRC(Bit_Chain * dat, long unsigned int start_address, 
+              long unsigned int size, uint16_t seed)
 {
   unsigned int calculated;
   unsigned int read;
   unsigned char res[2];
+  long unsigned int old_byte;
+  long unsigned int old_bit;
```
rurban: unneeded since it always must be 0. We always align dat for CRC calculations

src/common.c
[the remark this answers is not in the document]
rurban: We might need a seperate size array for all subtypes. 
I also need more for `work/dynapi`, like subtype of, size of vars, offset, but this will be seperate.

rurban:
MC/UMC always have max 4 bytes, not 5. It's either 2 or 4 bytes. See also the writer.
libdxfrw does the same.

DenisPryt:
I have a big [file](https://drive.google.com/file/d/13lK6PPtJk528APL-k3OBp-Mbu1XlzbXV/view?usp=sharing) with 5 bytes `offset`. See `read_2007_section_handles` function. 13'th iteration inside a `while (hdl_dat.byte - startpos < section_size)` loop

rurban:
Thanks for big.DWG. I can repro it now.

src/common.c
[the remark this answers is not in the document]
DenisPryt: Should I remove first commit (with composite type's sizes) form this pull request?

src/bits.c
DenisPryt: Why we need do this, if we have `start_address` parametr? Current `dat` bit and byte positions are not considered.

rurban: start_address is only the byte offset, with bit still 0.
What we need before that, is to do the alignment. You removed the alignment code.

src/bits.c
DenisPryt: There is a `start_address` parametr. It always align (directly bytes access by `dat->chain + start_address`). Input `dat` may not be not alligned in general case?

rurban:
So now there are only two issues missing:
* Use bit_check_CRC to verify object-map section: see `smoke/fixes`
* More accurate REPEAT_CHKCOUNT: add a minsize array for all subtype structs used in REPEAT, similar to all primitive BITCODE_ types. I have a similar array in `work/dynapi`

rurban:
I've looked at the Handle section crc check.
ok for everything but 2007, which uses a different 32bit crc check which we haven't added yet. 
It also has crc's for the uncompressed and compressed parts.

rurban:
I think of it is or obsolete/replaced by better code as of now. Esp. the new restricted streams catch overflows much better, and all CRC checks pass now.
If I forgot something not not please resubmit as new.

PR #66, opened 2018-12-07
Zero max_decomp_size fix

DenisPryt:
I have a file with `info->max_decomp_size == 0`. This leads to read memory acces error in `decompress_R2004_section`


rurban:
Merged into master with 0a7237c5feb50dc435e9751a62dde9a8f51fdb40

PR #67, opened 2018-12-07
Fix unknown class free problem

DenisPryt:
```
int dwg_decode_add_object() {
   Dwg_Object *obj;
   switch (obj->type) {
   case DWG_TYPE_VBA_PROJECT:
      error = DWG_ERR_UNHANDLEDCLASS;
      break;
   }
   // obj->tio.object is NULL
}

dwg_free_UNKNOWN_OBJ() {
   free(obj->tio.object->tio.token);
}
```


rurban:
thanks, looks good.
I'll still have to fix the new FreeBSD smoker

PR #72, opened 2018-12-12
Revert dat position in dwg_decode_add_object in all return cases.

DenisPryt:
dat position must be reverted. See `decode_R13_R2000` Object-map, section 2. 
In `while (dat->byte - startpos < section_size)` loop.


rurban:
Thanks, merged

PR #100, opened 2019-03-22
Misc. typos

luzpaz:
Found via `codespell`


rurban:
Nice, have to check it out

PR #213, opened 2020-03-24
fixed bug when out geojson to stdout

huangtao:
When i use dwgread a dwg file and destination  is std like: dwgread -O GeoJSON demo.dwg. I found the last 'Feature' has a COMMA,beacuse NOCOMMA(use fseek invalid on stdout).So i modify code to fixed this bug when the destination is  stdout.


huangtao:
Please forgive me for my poor English, ha ha ! 
To keep the original design (each array item ends with a comma), I put the need for a comma at the beginning of the feture if it is the first element of the array, output "{n", otherwise output ",{n",so emm SECOND HASH :).Maybe NOTFIRSTHASH is suitable.

rurban:
Ah, "second". I thought of "secure". I'll rewrite it a bit, but thanks for the idea!

huangtao:
Good job!

PR #215, opened 2020-03-25
Dockerfile

yossefaz:
Hi,

**Pull Request purpose**

I created a Dockerfile that can generate a stable image for using libreDWG.
I've tested this image and it works with different kind of DWG files.

**What I've done ?** 

I began from a python2.7 because test suites of libreDWG are written with this technology.

For libxml, i had many issues by installing this library and the only way i managed it was by compiling the libxml directly from the source code.

Then i build a multistage to get a thin and deployable image :  i used debian slim and not alpine because i did not manage to link the binaries with alipne (ldconfig).

**Go further**

In one of my project (https://github.com/yossefaz/dwg_transformer/tree/master/transformer) i use this image with an interface (a program in Golang) thant can receive request from a Queue to convert DWG files in different format.

I could also write a more generic program like mine, in order to expose all libreDWG commands to the Docker image. In this way users could interact directly out of the container to convert any kind of DWG : this would allow to run libreDWG in Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,41 @@
+############################
+# STEP 1 build executable binary
+############################
+
+FROM python:2.7.17-buster AS extracting
+LABEL maintainer="azoulayos@protonmail.com"
+RUN apt-get update && apt-get upgrade -y &&\
+    apt-get install -y git autoconf libtool swig texinfo build-essential gcc python-libxml2 && \
+    LIBXML2VER=2.9.1 && \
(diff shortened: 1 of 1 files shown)


yossefaz:
Okay : CodeFactor did not pass.
I will write and test with CodeFactor suggestion and send another pull request

rurban:
You can also skip the python xmltest. It's of not much use anyway than testing your local python installation. 

And most of our smokers are running with python 3 by now.

PR #216, opened 2020-03-25
Dockerfile

yossefaz:
Hi,

Pull Request purpose

I created a Dockerfile that can generate a stable image for using libreDWG.
I've tested this image and it works with different kind of DWG files.

What I've done ?

I began from a python2.7 because test suites of libreDWG are written with this technology.

For libxml, i had many issues by installing this library and the only way i managed it was by compiling the libxml directly from the source code.

Then i build a multistage to get a thin and deployable image : i used debian slim and not alpine because i did not manage to link the binaries with alipne (ldconfig).

Go further

In one of my project (https://github.com/yossefaz/dwg_transformer/tree/master/transformer) i use this image with an interface (a program in Golang) thant can receive request from a Queue to convert DWG files in different format.

I could also write a more generic program like mine, in order to expose all libreDWG commands to the Docker image. In this way users could interact directly out of the container to convert any kind of DWG : this would allow to run libreDWG in Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,41 @@
+############################
+# STEP 1 build executable binary
+############################
+
+FROM python:2.7.17-buster AS extracting
+LABEL maintainer="azoulayos@protonmail.com"
+RUN apt-get update && apt-get upgrade -y &&\
+    apt-get install -y git autoconf libtool swig texinfo build-essential gcc python-libxml2 && \
+    LIBXML2VER=2.9.1 && \
+    mkdir libxmlInstall && cd libxmlInstall && \
+    wget ftp://xmlsoft.org/libxml2/libxml2-$LIBXML2VER.tar.gz && \
+    tar xf libxml2-$LIBXML2VER.tar.gz && \
+    cd libxml2-$LIBXML2VER/ && \
+    ./configure && \
+    make && \
+    make install && \
+    cd /libxmlInstall && \
+    rm -rf gg libxml2-$LIBXML2VER.tar.gz libxml2-$LIBXML2VER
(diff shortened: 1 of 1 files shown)


rurban:
Still python 2.7?
Don't you need to upgrade sooner or later?

yossefaz:
> Still python 2.7?
> Don't you need to upgrade sooner or later?

I did not dee your previous comment about updrading python. So i will try to do so and post another pull request.

rurban:
I don'tt care much if python 2 or 3. You can choose, because you are the first docker user. Should be convenient to you.
But sooner or later you'll need to upgrade.

yossefaz:
Okay, for the moment i did not manage to build a Dockerfile with python3.
But the python2 image works perfectly.
Maybe we can merge it for now on master and i will continue to work on python3.

Travis did not pass...it's weird because it is related to an issue on the main code that i did not change.

yossefaz:
I manage to build a Dockerfile with python3
I will create a new (and hope last) pull request.

rurban:
Isn't it possible to force push to a PR?

yossefaz:
what do you mean by force push to a pull request ?

rurban:
There's not need to close, and open a new pull request on changes. just push -f to your branch, and the pull request gets updated automatically.

yossefaz:
ohhh okay. I did not knew that. Sorry. I will do in this way next time.

yossefaz:
I've already opened a new one. Is it okay for this time ?

rurban:
Sure. I'm just a bit busy with the json refactoring for stdout, but looks much better than the geojson approach.

PR #232, opened 2020-05-15
Fix 1.4 version

michal-josef-spacek:
DWG magic for this version is "AC1.40", not "AC1.4".


rurban:
Thanks. Do have such an example dwg also by chance? I'd love to have such an old DWG

PR #316, opened 2021-02-22
Added initial fuzzer for OSS-Fuzz integration.

DavidKorczynski:
Hi Maintainers,

I have worked a bit on setting fuzzing up for libredwg by way of OSS-Fuzz. Essentially, OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects, and it would be great to have libredwg integrated. The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. This is not a "hard" requirement in that no one enforces this and the main point is if bugs are not fixed then it is a waste of resources to run the fuzzers, which we would like to avoid.

In this PR https://github.com/google/oss-fuzz/pull/5226 I have done exactly that, namely created the necessary logic from an OSS-Fuzz perspective to integrate libredwg. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

At the moment I stored the fuzzers in the Google OSS-Fuzz repository, but we can move them up here to make maintenance easier.


rurban:
Thanks, very nice. It is maintained, sure.

The google maintainer emails are:
```
Reini Urban <reini.urban@gmail.com>
```

Did you see the special fuzzer binary `examples/dwgfuzz`? This is for afl, but libfuzzer from llvm is very similar to the in-memory fuzzing with afl++-clang, which is the default.

Your initial attempt only does decode fuzzing, there are several more codepaths, like
indwg, indxf, outdxf, injson, outjson to cover.

DavidKorczynski:
> Thanks, very nice. It is maintained, sure.
> 
> The google maintainer emails are:
> 
> ```
> Reini Urban <reini.urban@gmail.com>
> ```
> 
> Did you see the special fuzzer binary `examples/dwgfuzz`? This is for afl (afl++ with persistent in-memory), but libfuzzer from llvm is very similar to the in-memory fuzzing with afl++-clang, which is the default. I'll also add honggfuzz integration as it's much better also.
> 
> Your initial attempt only does decode fuzzing, there are several more codepaths, like
> indwg, indxf, outdxf, injson, outjson to cover.

Sounds good, and yes my initial fuzzer was inspired by `dwgfuzz.c` but I started with limited logic to see if it works. It should be fairly easy extending to the other procedures.

How would you like to proceed? Should we merge this in as well as the stuff on the oss-fuzz side, and then I can set up the infrastructure so it's easy for you to simply add new fuzzers in this repo in a manner where they are run by OSS-Fuzz?

rurban (changes requested):
I can take it asis, but I rather prefer using our existing much better fuzzing framework examples/dwgfuzz, which works with afl++ and hongfuzz already and covers much more critical code paths. Supports persistent mode already.

I also handle fuzzer regressions here: https://github.com/LibreDWG/libredwg-fuzz so I can easily test it against all previous fuzzing failures.
What do you think?
Let's say start with this one, and in the meantime I'll try adding llvm libfuzzer support to dwgfuzz.c and switch over later

rurban (approved):

rurban:
I've added now a better `examples/llvmfuzz` and use it with
`-timeout=4000 -detect_leaks=0 -rss_limit_mb=24000 ../test/test-data/`

It autodetects dwg, dxf or json, and writes to dwg, dxf, json or geojson.
Already found a few new bugs, libfuzzer looks very good.

PR #377, opened 2021-06-11
Refactor Spec file for building RPM packages

tjanez:
Copr repo with packages for current Fedora releases:
https://copr.fedorainfracloud.org/coprs/tadej/libredwg/

### DIFF
--- a/libredwg.spec
+++ b/libredwg.spec
@@ -1,7 +1,7 @@
 # -*- sh -*-
 Name:           libredwg
 Version:        0.12.4
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        GNU C library and programs to read and write DWG files
 
 License:        GPLv3+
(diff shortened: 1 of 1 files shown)


rurban:
Looks good, thanks for the copr. Just the wildcard for the manfiles looks dangerous on uninstall, and the dwgadd example is terribly needed

tjanez:
> Looks good, thanks for the copr.

Thanks! And thanks for merging this.

> Just the wildcard for the manfiles looks dangerous on uninstall

It shouldn't be dangerous on uninstall since RPM only uninstalls the exact list of files that belong to the installed RPM package.

Sometimes, packagers avoid wildcard under `%files` to not unintentionally package some things that were (erroneously) installed and are not meant to be part of the package.

But I think in this case, it should be safe to assume all files matching:
```
%{_mandir}/man1/dwg*
%{_mandir}/man5/dwg*
%{_infodir}/LibreDWG.info*
```
should be included in the package.

> and the dwgadd example is terribly needed

He he :). I think we could include it, I would just prefer to put it to `/usr/share/libredwg` instead of directly to `/usr/share` since none of the packages put things directly to `/usr/share`.

```

rurban:
`/usr/share` oops, my mistake. will fix it asap

PR #405, opened 2022-02-13
Add AXISMODE and AXISUNIT

michal-josef-spacek:
Tested on DWG AC1001-AC1009.
AC2.10 has same but not working due issue in block sizes.


rurban:
Nice! And quite short, so we don't need the GNU copyright assignment

PR #410, opened 2022-02-17
GH#409: Fix OSMODE header variable

michal-josef-spacek:
From AC1.50 to AC1009 (old format) there is issue in place of OSMODE
variable.

BTW: FIELD_RS (unknown_12, 8); is TEXTSTYLE, but this is for another
issue, we have second TEXTSTYLE variable for different DWG format.

Values:
  osnap_modes:
    0: none
    1: endpoint
    2: midpoint
    4: center
    8: node
    16: quadrant
    32: intersection
    64: insertion
    128: perpendicular
    256: tangent
    512: nearest


rurban (approved):
Ah, my old favorite variable!

rurban:
Note, that I've disabled travis-ci now. I didnt pay them, and dont want to

PR #416, opened 2022-02-19
Decode variables in AC1.50

michal-josef-spacek:
These variables are valid from AC1.50 to AC1009


rurban (approved):

michal-josef-spacek:
@rurban How is working output like 16b0c8b126d94463a9fb75aa825e45d7278f77db..4a16f42c557eb35e1ab67bf69ff85228c004d65b ? Is this code autogenerated or written?

rurban:
make regen-dynapi

michal-josef-spacek:
@rurban 
When i am using configure then `make regen-dynapi` causes error:
```
woman:~libredwg> make regen-dynapi
/usr/bin/perl ./src/gen-dynapi.pl
using gcc with /usr/lib/gcc/x86_64-linux-gnu/8/include
Cannot find input file '../include/dwg.h' at ./src/gen-dynapi.pl line 82.
make: [Makefile:1244: regen-dynapi] Chyba 2 (ignorováno)
```

rurban:
I only tested it via ../configure in a build dir.
Otherwise you need a special perl call, and a Makefile fix

rurban:
Already merged

PR #430, opened 2022-04-12
Add missing informations about entities in pre r13 releases

michal-josef-spacek:


src/dwg.spec
```diff
@@ -802,9 +802,7 @@ DWG_ENTITY (MINSERT)
 
 DWG_ENTITY_END
 
-/*
-// 9 was TRACE. r10-r11 only
```
rurban, line 806: I'd like to keep the TRACE comment there

rurban (changes requested):
just the TRACE comment.
Haven't you found much more infos with your Kaitai parsers? I'm just playing with them.

rurban:
Na, we are good. I'll rebase it onto mine now

src/dwg.spec
michal-josef-spacek: I don't think that comment is true. But no problem with it.

michal-josef-spacek:
> BTW: I'm just now fixing the preR13 parsing, based on your ksy files. Looks much better now.

ok, thanks. I have some changes too.

You could look to this commit (https://github.com/michal-josef-spacek/libredwg/commit/d69ff8a13b9fdb5555c404fef109ce5562d7f1ff), but still in progress. My plan was prepare PRs for things which i know and go step by step.

src/dwg.spec
michal-josef-spacek: Ah entity id 9 for trace is in <AC1.50 - look https://github.com/michal-josef-spacek/CAD-Format-DWG-AC1_40/blob/main/dwg_ac1_40.ksy

FYI: format of DWG to <AC1.50 is one, from AC1.50 to AC1009 second and >AC1009 third. We could add decoding section for it. I know everything in AC1.40 instead of structure in 0x01d0-0x01d7 (i have 8 byte float, but i don't think that it is - converted to format  in AC1.50 and above. Seems as some structure).

michal-josef-spacek: uff. I was looking to it. And trace is id 9 in all versions to <=AC1009. And i don't understand sense of this comment. This is comment for 3DLINE.

rurban:
Looks good. I've rebased it onto `smoke/versions`

michal-josef-spacek:
@rurban great. I will look to it later. 

FYI: I commented my WIP branch, there are some issues for fix (https://github.com/michal-josef-spacek/libredwg/commit/d69ff8a13b9fdb5555c404fef109ce5562d7f1ff).

rurban:
Just VECTOR_CHKCOUNT is broken for you. Will look into it.

BTW: I'm using the local kaitai webide. The ksv tui alone is unusable for me.

rurban (approved):

rurban:
Looks good now in `smoke/preR13`, will merge soon.
I'm also thinking of encode support for those early versions, because they are so trivial.

rurban:
Merged

michal-josef-spacek:
@rurban There are issues in entities and tables still. I am working on it.

rurban:
@michal-josef-spacek I'll merge smoke/preR13 soon. even with the windows dupl. symbol problem. this can decide and encode many.

michal-josef-spacek:
@rurban please wait with merge. I found issue with header values yesterday. Some shift by one byte in some block. I need to look.

PR #431, opened 2022-04-13
Identify new pre R13 header variables

michal-josef-spacek:
Signed-off-by: Michal Josef Špaček <michal.josef.spacek@gmail.com>

### DIFF
--- a/src/header_variables_r11.spec
+++ b/src/header_variables_r11.spec
@@ -90,37 +90,29 @@
   if (dwg->header.num_header_vars <= 74)
      return 0;
   FIELD_RC (LIMCHECK, 70); //ok 1fa
-  //DEBUG_HERE //1fb
-  // just guessing. not in DWG, resp. not converted by ODA from r12 dxf to r12 dwg:
-  //   SNAPMODE, DRAGMODE, BLIPMODE, CHAMFERA, CHAMFERB,
-  //   COORDS, ATTDIA, ATTREQ, HANDLING, WORLDVIEW, SHADEDGE
-  FIELD_RC (unknown_10, 70);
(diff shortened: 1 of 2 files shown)


michal-josef-spacek:
@rurban update of variables. I tested it and values seems ok. 

Some unknown things:
* DIMBLK_T, DIMBLK1_T, DIMBLK2_T length 33?
* Need to fix HANDSEED
* I don't understand your guessing section. e.g. When i am looking you created DRAGMODE second time. All variables are in my change.
* I don't undestand PSLTSCALE, why is there? I haven't reproducer for this variable.

FYI: CHAMFERA is in some releases in DXF but not in DWG :-D (e.g. r2.18)
FYI: This number of variables in header - this is not related to precise release. Sometime it is related to "Advanced Drafting Extension", if i understand.

michal-josef-spacek:
@rurban i believe that section
```
PRE(R_2_0) // AC1.2 definitely
     return 0;
```
doesn't sense.
AC1.2 and AC1.40 has more diferences than this section. See https://github.com/michal-josef-spacek/CAD-Format-DWG-AC1_40/blob/main/dwg_ac1_40.ksy

src/header_variables_r11.spec
```diff
@@ -90,37 +90,29 @@
   if (dwg->header.num_header_vars <= 74)
      return 0;
   FIELD_RC (LIMCHECK, 70); //ok 1fa
-  //DEBUG_HERE //1fb
-  // just guessing. not in DWG, resp. not converted by ODA from r12 dxf to r12 dwg:
```
rurban, line 94: Better but I want to keep these comments of mine in

rurban (approved):

PR #434, opened 2022-04-19
Improve flags, options and point entity

michal-josef-spacek:


rurban:
Merged. 
But I still wamt to move thickness and extrusion back to the entities. Even if we have to copy these 4 lines into every entity spec.

rurban:
And I think I saw extruded points, but forgot with which version.

michal-josef-spacek:
> Merged. But I still wamt to move thickness and extrusion back to the entities. Even if we have to copy these 4 lines into every entity spec.

Thanks. Yes, that's posibility to copy to entity.
FYI: I saw big 10?? (dxf) block for vport (entity) before handling and pspace flag when i created line in paper space in R14 and export to R12 dwg. 

> And I think I saw extruded points, but forgot with which version.

In <=R11 doesn't exists extrude command.
In R14 is extrude command and is creating solid from e.g circle, rectangle, closed pline, etc.
Conversions to R12 dwg partially working in this case, but i have much of bugs to read still

PR #437, opened 2022-04-20
Smoke/pre r13 foo

michal-josef-spacek:
Fix PSLTSCALE and TREEDEPTH
Fix pre R_2_0 entities


rurban:
I think most of it is already in. Can you rebase, or should I check?

michal-josef-spacek:
@rurban I fixed conflicts

rurban:
I've rebased it, and added the needed dynapi and tests.

michal-josef-spacek:
@rurban great

rurban:
LOAD still fails the unit-test, will look at it tommorrow

PR #438, opened 2022-04-20
Add AC1.40 test DWG file

michal-josef-spacek:


rurban:
Will be in the #437 merge

rurban:
merged

PR #440, opened 2022-04-23
Add AC2.10 test DWG file

michal-josef-spacek:
With script definition


rurban:
can we put that it into a r2.10 dir please

michal-josef-spacek:
@rurban Changed and added polyline

rurban:
rebased and merged

PR #445, opened 2022-04-25
Fix preR13 shape entity

michal-josef-spacek:
There is no z point (is in common elevation) and extrusion


rurban:
Replaced by 71b386ab764f801325ec499b589e13e01c074487
Very similar but different

PR #448, opened 2022-04-26
Fix preR13 SHAPE and POLYLINE entities

michal-josef-spacek:


michal-josef-spacek:
@rurban FYI: Seems all AC2.10 entities are ok (which i am able to test for now).

rurban:
DXF 66 means has_children, but since they are INSERT/POLYLINE specific only, we named them has_attrib and has_vertex. 

DXF does not really export it anymore, since has_vertex is now always 1. When you delete all vertices, the POLYLINE parent gets deleted also.

michal-josef-spacek:
@rurban ok, i understand a sense of this 66 DXF, but i don't know if exist in DWG file or not.
And don't know where is 70 field.

PLINE_CLOSED in DXF: 66 = 1, 70 = 1
PLINE_SPLINE in DXF: 66 = 1, 70 = 8

rurban:
already merged

PR #450, opened 2022-04-26
Fix dwgadd build without write support

michal-josef-spacek:


rurban:
bah, removing the example for the poor guys without write? 
this could serve as documentation example to upgrade to write support.

michal-josef-spacek:
Heh, i only added this if for all things which are related to dwgadd :-)

rurban:
Na, I rather want to keep it

PR #451, opened 2022-04-26
Fix preR13 3DFACE entity

michal-josef-spacek:


rurban (approved):

rurban:
I'll merge it into master instead. work/preR13 is for tables refactoring (dxf, json) only, and needs some time.

rurban:
Merged into master instead

PR #452, opened 2022-04-27
Fix entities and block entities decoding

michal-josef-spacek:


src/decode_r11.c
```diff
@@ -799,10 +804,11 @@ decode_preR13 (Bit_Chain *restrict dat, Dwg_Data *restrict dwg)
...
       dat->byte = blocks_start;
     }
-  // don't just decode the num_blocks (i.e. 1 the BLOCK ent),
-  // but the num_entities minus the already decoded entities.
-  //if (dwg->header.section[SECTION_BLOCK].number)
-  num_entities -= dwg->num_entities; // minus the already decoded modelspace ents
+  // this is number of block entities, not count of blocks.
+  num_entities = 0;
+  if (R_11) {
+    blocks_end -= 32; // crc size
```
rurban, line 810: that cannot be. a crc16 has 2 byte, a sentinel 16. I'll check what is there.

michal-josef-spacek: @rurban There are 32 bytes, which are unknown for me for now. Sorry for crc word.

FYI actually i have this code:
```
-  // don't just decode the num_blocks (i.e. 1 the BLOCK ent),
-  // but the num_entities minus the already decoded entities.
-  //if (dwg->header.section[SECTION_BLOCK].number)
-  num_entities -= dwg->num_entities; // minus the already decoded modelspace ents
+  num_entities = 0;
+  if (! blocks_end) {
+    blocks_end = blocks_start + (blocks_size & 0x3FFFFFFF);
+  }
+  SINCE (R_11) {
+    blocks_end -= 32; // crc size
+  }
```

PR #455, opened 2022-04-28
Fix preR13 line entity

michal-josef-spacek:


michal-josef-spacek:
@rurban sorry, that's only rebase, same as you probably.

rurban:
I"v merged it into master instead. work/ branches are unstable

rurban (approved):
Ok for master

rurban:
Merged into master

PR #458, opened 2022-05-02
Align with r11 header end

michal-josef-spacek:


src/header_variables_r11.spec
```diff
@@ -323,3 +323,7 @@
     return 0;
   FIELD_CAST (VISRETAIN, RS, B, 70); //ok
 
+  /* TODO Unknown 18 bytes */
+  DEBUG_HERE //0x6bd-0x6ce
+  dat->byte = 0x6cf;
```
rurban, line 328: I think, we should better store the unknown slack in some buffer. For encode

rurban (changes requested):

src/header_variables_r11.spec
michal-josef-spacek: Hm, this is probably not part of header.
I have two versions of DWG, which have 204 and 205 header variables. And difference is VISRETAIN value.

rurban: I suspect it's a similar kind of sentinel, some magic, unique bytes after the headers, and before the entities

rurban:
Replaced by something better

michal-josef-spacek:
@rurban ah, interesting. Small fix for this https://github.com/LibreDWG/libredwg/pull/464

PR #461, opened 2022-05-03
Fix trailing bytes for table check

michal-josef-spacek:


rurban (approved):
Sure

rurban:
merged and revised in the smoke branch

PR #462, opened 2022-05-04
Fix shape entity handle

michal-josef-spacek:


rurban:
That's already in since 71b386ab764f801325ec499b589e13e01c074487

michal-josef-spacek:
@rurban Ahh, not in preR13-iterators branch

PR #464, opened 2022-05-05
Fix R11 header sentiel

michal-josef-spacek:
There are multiple R11 header variants (204 and 205 header variables)


rurban:
merged

PR #477, opened 2022-05-31
Fix style in ATTRIB entity

michal-josef-spacek:


michal-josef-spacek:
Similar as https://github.com/LibreDWG/libredwg/pull/476

rurban:
Merged, thanks

PR #480, opened 2022-05-31
Fix layer linetype index in preR13

michal-josef-spacek:


michal-josef-spacek:
Two bytes for index, tested in AC2.10 on 258 linetypes

PR #500, opened 2022-06-27
Fix various typos

luzpaz:
Found via `codespell -q 3 -S ChangeLog -L 2rd,ba,childs,daa,ded,fo,oce,ro,siz`


rurban (approved):
Ouch, so many!
Thanks a lot

PR #506, opened 2022-07-28
Fixes Stack-buffer-overflow in decompress_rNUMBER

sashashura:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44432

`sizeof (r2007_file_header) + 1` is passed in `read_file_header` which leads to off by one write buffer overflow.


rurban:
Hmm, why did I do that?? thanks!

rurban:
The fix changes the off-by-one only in 1 of 3 places. I harmonized it and rather fixed the out-of-bounds checks in 83a71b19797ac78eba855faa7df6908d55d2f65d

Issue #471, opened 2022-05-13
Issue with entity parsing in pre2.10


michal-josef-spacek:
There is something wrong with entity parsing in work/preR13 branch.
See example: 
[shape.tar.gz](https://github.com/LibreDWG/libredwg/files/8691343/shape.tar.gz)
There are two entities:
* LOAD
* SHAPE

Parse only first entity, because bad number "num" in src/decode.c - https://github.com/LibreDWG/libredwg/blob/09607bb52479d69356d07346e795fd98b7b867b9/src/decode.c#L5653
There is num = 8 on begin
and then break in:
```
        if (num + 1 >= dwg->header_vars.numentities)
          break;
```
so parse only first entity

michal-josef-spacek:
It's working now, thanks

PR #533, opened 2022-12-09
Fix parsing of AutoCAD 1.4 entities

michal-josef-spacek:
num is index of actual object

Fixing https://github.com/LibreDWG/libredwg/issues/471


rurban:
Just fixed the failing test for LOAD

rurban:
Merged

PR #534, opened 2022-12-11
Fix writing of 1.4 release DWG files

michal-josef-spacek:
We are setting entity type for proper version and <R11 has different types.

Fix going over all objects.


rurban (approved):
Ah, I see. Looks good

PR #537, opened 2022-12-15
Support MC0_0

michal-josef-spacek:


michal-josef-spacek:
FYI: I have MC0.0, AC1.2, AC1.40, AC1.50, AC1001, AC1002, AC1003, AC1004 .... DWG files. 
And looked at AutoCAD release 3 (1.3) and release 4 (1.4) distributions, where aren't AC1.3 files.

PR #543, opened 2023-01-02
Add extrusion to insert entity

michal-josef-spacek:


rurban:
LGTM, but I'll rebase it on top of smoke/flags

rurban:
merged

PR #545, opened 2023-01-03
Fix polyline mesh entity for preR13

michal-josef-spacek:


rurban:
better (fixes num_m_verts, num_n_verts), but still 4 byte too many. m/n_density behind opts also. curvetype also

rurban:
merged with changes

PR #546, opened 2023-01-03
Improve logging

michal-josef-spacek:


include/dwg.h
```diff
@@ -8064,6 +8064,17 @@ enum {
   FLAG_R11_HAS_ATTRIBS = 128,
 };
 
+enum {
```
rurban, line 8067: Those test generation flags are actually for all versions, not just R11

include/dwg.h
```diff
@@ -8075,6 +8075,308 @@ enum {
   OPTS_R11_XDATA = 128, //?
 };
 
+enum {
+  OPTS_R11__3DFACE_HAS_Z_FIRST = 1,
+  OPTS_R11__3DFACE_HAS_Z_SECOND = 2,
+  OPTS_R11__3DFACE_HAS_Z_THIRD = 4,
+  OPTS_R11__3DFACE_HAS_Z_FOURTH = 8,
+  OPTS_R11__3DFACE_UNKNOWN_16 = 16,
```
rurban, line 8116: I prefer using the R11 types instead, esp. for DIMENSION, POLYLINE and VERTEX later.
And I prefer to skip all the unused flags.

rurban:
With minor changes as smoke/flags. Looks good now

include/dwg.h
michal-josef-spacek: I understand. First version :-)

include/dwg.h
michal-josef-spacek: DIMENSION, POLYLINE, VERTEX are commons entities, we need somehow unify.

btw: These unused flags are good to check existing flags in real DWG files.

rurban:
Merged with changes

include/dwg.h
rurban: I see the reason why, but I rather keep only used flags, and maybe warn when some unused flag appears out of the blue.

michal-josef-spacek: If you have a way how to print that flag that is not defined, that's ok for me.

rurban: See 23b1fc8c5d245c73bca299cf958d983e0208e93f
cleanup unused r11 opts and flags

PR #547, opened 2023-01-03
Fix bigfont_file in style table

michal-josef-spacek:


rurban:
LGTM, merged into smoke/fuzz

rurban (approved):

PR #550, opened 2023-01-03
Identify vertex flags

michal-josef-spacek:


rurban:
Almost, just not the MESH names

michal-josef-spacek:
ok, thanks

PR #552, opened 2023-01-03
Fix _3DFACE entity

michal-josef-spacek:


rurban:
Merged into smoke/flags

PR #555, opened 2023-01-04
Log block flag

michal-josef-spacek:


rurban:
I'd rather use the shortened names from r2000+

michal-josef-spacek:
> I'd rather use the shortened names from r2000+

I am ok with it. I created them from the DXF reference manual.

rurban:
Merged the fixed up version into smoke/flags

PR #559, opened 2023-01-04
Fix blocks_end

michal-josef-spacek:


src/decode_r11.c
```diff
@@ -728,6 +728,10 @@ decode_preR13 (Bit_Chain *restrict dat, Dwg_Data *restrict dwg)
         LOG_TRACE ("blocks_size: " FORMAT_RL " [RL]\n", blocks_size);
       }
     blocks_end = bit_read_RL (dat);
+    if (blocks_end == 0 && blocks_size != 0)
```
rurban, line 731: This rather denotes an empty BLOCKS section. I've fixed up the logic below which processes the BLOCKS

michal-josef-spacek: ok, thanks

PR #560, opened 2023-01-04
Log attrib entity flags

michal-josef-spacek:


rurban (changes requested):
I'll rather keep the spec comments. The logs are rarely used

rurban:
Merged the fixed up version into smoke/flags

PR #561, opened 2023-01-05
Identify text extrusion

michal-josef-spacek:


src/dwg.spec
```diff
@@ -50,6 +50,8 @@ DWG_ENTITY (TEXT)
     }
     if (R11OPTS (64))
       FIELD_2RD (alignment_pt, 11);
+    if (R11OPTS (128))
+      FIELD_3RD	(extrusion, 210);
```
rurban, line 54: There was a tab in there, fixed

PR #562, opened 2023-01-05
Log table flags

michal-josef-spacek:


rurban (changes requested):
I'd rather seperate the common table flags from the specific ones

michal-josef-spacek:
Something like this?

PR #563, opened 2023-01-05
Log LWPOLYLINE flag

michal-josef-spacek:


rurban:
Nope, 1024 vertexidcount is missing

michal-josef-spacek:
I have DWG file with 256, but i don't know what is it :-)

rurban:
Merged, slightly changed

PR #570, opened 2023-01-11
Fix layer table

michal-josef-spacek:


src/decode_r11.c
```diff
@@ -380,18 +380,12 @@ decode_preR13_section (Dwg_Section_Type_r11 id, Bit_Chain *restrict dat,
...
           FIELD_CMC (color, 62); // off if negative
-          PRE (R_11)
-          {
-            FIELD_HANDLE (ltype, 2, 6);
-            if (tbl->size == 38)
-              FIELD_RC (flag0, 0);
-          }
-          LATER_VERSIONS
-          {
-            FIELD_RS (linewt, 370);
```
rurban, line 392: no r11 linewt?

michal-josef-spacek: I don't find it.
In Autocad R14 (> R11) isn't `Lineweight enum value` as I understand.

michal-josef-spacek: Ahh, Lineweight is the first time in Autocad 2000 and there is no export to R11.

PR #577, opened 2023-01-15
Fix logs in free

michal-josef-spacek:


rurban:
I'd like to see all these nice flags on dxf export also. Just FREE should be excluded.

michal-josef-spacek:
> I'd like to see all these nice flags on dxf export also. Just FREE should be excluded.

good point

PR #583, opened 2023-01-18
Fix preR13 dimstyle

michal-josef-spacek:


michal-josef-spacek:
TODO: Other dimension types

rurban:
> TODO: Other dimension types

I saw it for all types with r11

michal-josef-spacek:
> > TODO: Other dimension types
> 
> I saw it for all types with r11

I mean src/dwg.spec file
dimstyle is for all dimension types in R10/R11

PR #593, opened 2023-01-20
Add r9 test files

michal-josef-spacek:


rurban:
They need to be added to the Makefile.am also. Will do

PR #598, opened 2023-01-22
Identify flag_3d in preR13 header

michal-josef-spacek:
It's same variable as flag_3d in VIEW table.
Not saved to DXF


rurban:
I see that you've fixed your gen-dynapi.pl now, good

michal-josef-spacek:
> I see that you've fixed your gen-dynapi.pl now, good

No. Sorry, I don't know what with it. I only changed evident things by hand.

michal-josef-spacek:
plus ```make regen-dynapi```

PR #602, opened 2023-01-23
Regen unknown_59 header variable

michal-josef-spacek:


rurban:
I've merged a better fix for this

PR #604, opened 2023-01-24
Fix gen-dynapi.pl

michal-josef-spacek:
Debian has includes in arch dependent directory too


rurban:
Not just Debian, everybody has. But your patch looks OK, will test

PR #605, opened 2023-01-24
Identify UCSNAME in preR13

michal-josef-spacek:
This is index to UCS table


rurban:
Very good!

michal-josef-spacek:
Heh, another is PUCSNAME, preparing update

PR #612, opened 2023-01-28
Some MC0.0 support

michal-josef-spacek:


rurban (approved):

michal-josef-spacek:
dwgadd is failing still

michal-josef-spacek:
Issue is with free on insert entity.
Root cause is:
```
  if (dwg->header.version < R_2_0b)
    _obj->block_name = (BITCODE_TV)name;
```
I don't understand why.

src/dwg_api.c
```diff
@@ -23610,6 +23610,11 @@ dwg_add_DIMENSION_ALIGNED (Dwg_Object_BLOCK_HEADER *restrict blkhdr,
                            const dwg_point_3d *restrict text_midpt)
 {
   API_ADD_ENTITY (DIMENSION_ALIGNED);
+  if (dwg->header.version <= R_2_22)
+    {
+      LOG_ERROR ("Invalid entity %s <R7", "DIMENSION_ALIGNED")
```
rurban, line 23620: please use lowercase and the public revisions, as in common.c: dwg_versions[]
"<r2.22"

rurban (changes requested):

src/dwg_api.c
michal-josef-spacek: No problem, but all version comparisons in this file are like this one.

rurban:
Done in smoke/some_mc0_support

michal-josef-spacek:
Thanks

PR #620, opened 2023-01-31
Fix VPORT table

michal-josef-spacek:

PR #621, opened 2023-01-31
Fix UCS table

michal-josef-spacek:


rurban:
I need to test this, as I had the very same idea, but found contradicting examples

michal-josef-spacek:
> I need to test this, as I had the very same idea, but found contradicting examples

I think that's ok - based on fix with used in https://github.com/LibreDWG/libredwg/pull/620

michal-josef-spacek:
> > I need to test this, as I had the very same idea, but found contradicting examples
> 
> I think that's ok - based on fix with used in #620

Ah, no.. Missing used in UCS.. I will fix

rurban:
tested ok

PR #624, opened 2023-02-01
Add start_width and end_width to POLYLINE_3D entity

michal-josef-spacek:


rurban:
Oh my! I guess this comes from their unified POLYLINE implementation preR13. So they had to carry all this ballast along.

rurban:
Merged

PR #627, opened 2023-02-02
Some fixes

michal-josef-spacek:


rurban (approved):
just fixed the very same BTW

rurban:
I think I merged now all

PR #630, opened 2023-02-08
Fix sections

michal-josef-spacek:


rurban:
See `work/fix_sections`

rurban:
Merged with my fixups

Issue #548, opened 2023-01-03
Issue with Polyline in AC1003


michal-josef-spacek:
[fit_curve_pline.scr.gz](https://github.com/LibreDWG/libredwg/files/10336468/fit_curve_pline.scr.gz)
[fit_curve_pline.dxf.gz](https://github.com/LibreDWG/libredwg/files/10336469/fit_curve_pline.dxf.gz)
[fit_curve_pline.dwg.gz](https://github.com/LibreDWG/libredwg/files/10336470/fit_curve_pline.dwg.gz)

Something wierd:
```
entities: (0x3d5-0x431 (7), size 92, 0x0)
==========================================
type: 18 [RCd]
Add entity POLYLINE_2D [8] Decode entity POLYLINE_2D
===========================
Entity number: 8, Type: 18, Addr: 3d5
flag_r11: 0x0 [RC 0]
size: 8 [RS]
layer: 146 [H(RSd) 8]
opts_r11: 0x8000 [RSx 0]
```
1) layer: 146, sic.
2) opts_r11: 0x8000

Second polyline:
```
type: 18 [RCd]
Add entity POLYLINE_2D [9] Decode entity POLYLINE_2D
===========================
Entity number: 9, Type: 18, Addr: 3dd
flag_r11: 0x0 [RC 0]
size: 24 [RS]
layer: 73 [H(RSd) 8]
opts_r11: 0x8000 [RSx 0]

ERROR: offset 16
```
Different size, same type.

michal-josef-spacek:
R11 output:

[fit_curve_pline_r11.dxf.gz](https://github.com/LibreDWG/libredwg/files/10337426/fit_curve_pline_r11.dxf.gz)
[fit_curve_pline_r11.dwg.gz](https://github.com/LibreDWG/libredwg/files/10337427/fit_curve_pline_r11.dwg.gz)

Same

michal-josef-spacek:
R14 output exported as R11:

[fit_curve_pline_r14_exported_as_r11.dwg.gz](https://github.com/LibreDWG/libredwg/files/10337461/fit_curve_pline_r14_exported_as_r11.dwg.gz)

Different:
```
entities: (0x6cf-0x79a (0), size 203, 0x0)
==========================================
type: 19 [RCd]
Add entity POLYLINE_2D [13] Decode entity POLYLINE_2D
===========================
Entity number: 13, Type: 19, Addr: 6cf
flag_r11: 0xa0 [RC 0]
[...]

PR #634, opened 2023-02-09
Fix polyline 0x8000 flag

michal-josef-spacek:
GH #548
Flag mean, that entity is replaced by entity in extra entity section. There are unknown data sometime


rurban:
It could be the calculated curve fit pts

michal-josef-spacek:
> It could be the calculated curve fit pts

Yes, I think too.

michal-josef-spacek:
When I open this kind of file and save it again, entities were replaced.

rurban:
You deleted the code to get the 2RD curve_fit pts. NOK

PR #641, opened 2023-02-14
Remove the type cast not supported by VS2019

gongxy-sh:


rurban:
There are hundreds of them.
Does it build at all with VS2019?

gongxy-sh:
encode.c/2  decode.c/11 dwg_api.c/1 and disable logging.h by config.h

gongxy-sh:
Should I open one PR per commit or can I merge multiple commits in one PR? CI takes too much time.

rurban:
All in one commit. But this is not trivial anymore , and needs an FSF Contribution assignment. See CONTRIBUTING.md

rurban:
Also, VS2019 is not C99 compatible, so I don't think we will be able to add support it. mingw gcc,  ICC or clang-cl are much better

src/logging.h
```diff
@@ -96,11 +96,11 @@
     }
 #endif
 
-#define LOG_INFO(args...) LOG (INFO, args)
-#define LOG_TRACE(args...) LOG (TRACE, args)
-#define LOG_HANDLE(args...) LOG (HANDLE, args)
-#define LOG_INSANE(args...) LOG (INSANE, args)
-#define LOG_ALL(args...) LOG (ALL, args)
+#define LOG_INFO(...) LOG (INFO, ##__VA_ARGS__)
```
rurban, line 99: I see. This probably needs a _MSC_VER switch

rurban (commented):
Fair enough. Not that bad.

I'll try to setup a CI build for VS2019. The project files should not be needed, cmake would do that.

rurban:
I've added a work/VS2019 branch for non-standard msvc support, via clang-cl. (and I also tested other smaller compilers, such as tcc and chibicc)

It is possible to test it locally now.

gongxy-sh:
OK. 

sed -i 's/args\.\.\./.../' ../src/logging.h
sed -i 's/args/##__VA_ARGS__/' ../src/logging.h

sed -i 's/const unsigned long size = sizeof (Dwg_R2004_Header);/#define R2004_HEADER_SIZE sizeof (Dwg_R2004_Header)/' ../src/decode.c
sed -i 's/BITCODE_RC decrypted_data\[size\];/BITCODE_RC decrypted_data[R2004_HEADER_SIZE];/' ../src/decode.c
sed -i 's/decrypted_header_dat.size = size;/decrypted_header_dat.size = R2004_HEADER_SIZE;/' ../src/decode.c
sed -i 's/file_dat->size < size/file_dat->size < R2004_HEADER_SIZE/' ../src/decode.c
sed -i 's/size + 0x80, size);/R2004_HEADER_SIZE + 0x80, R2004_HEADER_SIZE);/' ../src/decode.c
sed -i 's/\], (int)size);/], (int)R2004_HEADER_SIZE);/' ../src/decode.c
sed -i 's/\], size);/], R2004_HEADER_SIZE);/' ../src/decode.c
sed -i '3781d' ../src/decode.c
sed -i '3781i                  //ACAD_MLEADERVER' ../src/decode.c
sed -i 's/\*dp, \*sp, \*dp0, \*sp0;/*dp=0, *sp=0, *dp0=0, *sp0=0;/' ../src/decode.c

sed -i '34i#endif' ../src/dwg.c
sed -i '33i#ifdef HAVE_LIBGEN_H' ../src/dwg.c

sed -i 's/obj->type = -(obj->type);/obj->type = -(long)(obj->type);/' ../src/dwg_api.c

sed -i 's/(BITCODE_TIMEBLL)_obj->nam/_obj->nam/' ../src/encode.c
sed -i 's/(BITCODE_TIMERLL)_obj->nam/_obj->nam/' ../src/encode.c

 Then code is VS2019/GCC compatible.

PR #643, opened 2023-02-15
Auxiliary functions for dwg R/W.

gongxy-sh:
dwg_new_Document0/dwg_free_Document for DWG reader.
dwg_free_Document for DWG writer.

### DIFF
--- a/include/dwg.h
+++ b/include/dwg.h
@@ -9574,7 +9574,7 @@ typedef struct _dwg_struct
   BITCODE_H *acis_sab_hdl;
   unsigned long next_hdl; // for add_document handle holes
 
-  size_t num_ordered_object_refs; /*!< number of ordered object refs */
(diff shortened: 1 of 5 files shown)


rurban (approved):
Ok, fair enough.
dwg_free is exported, but this pleases our VBA naming style.

gongxy-sh:
dwg_free_Document = dwg_free + free.

gongxy-sh:
dwg_free_Document = dwg_free + free.

gongxy-sh:
for reader
Dwg_Data *dwg  = dwg_new_Docment0(...);
dwg_read_file(...)
...
dwg_free_Document(dwg);

for writer
Dwg_Data *dwg  = dwg_new_Docment(...);
...
dwg_write_file(...)
dwg_free_Document(dwg);

PR #644, opened 2023-02-15
Speed up DWG writer

gongxy-sh:
Before：
100000 entities used 00:03:34
300000 entities used 00:33:53

After:
100000 entities used 00:00:02
300000 entities used sed 00:00:23

### DIFF
--- a/src/dwg.c
+++ b/src/dwg.c
@@ -72,9 +72,19 @@ dwg_find_tablehandle_silent (Dwg_Data *restrict dwg, const char *restrict name,
 // used in encode.c
 void set_handle_size (Dwg_Handle *restrict hdl);
 
-void dwg_ordered_ref_appended (Dwg_Data *restrict dwg, Dwg_Object_Ref *ref);
-Dwg_Object_Ref *dwg_ordered_ref_find (Dwg_Data *restrict dwg, const BITCODE_RC code,
-                                      const unsigned long absref);
+static void ordered_ref_add (Dwg_Data *restrict dwg, Dwg_Object_Ref *ref);
+
+static Dwg_Object_Ref *ordered_ref_find (Dwg_Data *restrict dwg,
+                                         const BITCODE_RC code,
+                                         const unsigned long absref);
+
+static int ordered_ref_cmp (const Dwg_Object_Ref *key,
+                            const Dwg_Object_Ref **ppR);
+
+static void *bsearch_ex (const void *key, const void *base, size_t num,
+                         size_t width,
+                         int (*compare) (const void *item1, const void *item2),
+                         const void **ppBefore);
 
 /*------------------------------------------------------------------------------
  * Public functions
@@ -2037,7 +2047,7 @@ dwg_add_handleref (Dwg_Data *restrict dwg, const BITCODE_RC code,
         }
         */
       //change lsearch to bsearch
-      Dwg_Object_Ref *reff = dwg_ordered_ref_find (dwg, code, absref);
+      Dwg_Object_Ref *reff = ordered_ref_find (dwg, code, absref);
       if (NULL != reff)
         {
           return reff;
@@ -2049,7 +2059,7 @@ dwg_add_handleref (Dwg_Data *restrict dwg, const BITCODE_RC code,
   ref->absolute_ref = absref;
   ref->obj = NULL;
   // fill ->obj later
-  dwg_ordered_ref_appended (dwg, ref);
+  ordered_ref_add (dwg, ref);
   return ref;
 }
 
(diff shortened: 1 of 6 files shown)


gongxy-sh:
Main code as 
Dwg_Data* pDoc = dwg_new_Document(R_2000, 0, 0);
  Dwg_Object* poMSpace = dwg_model_space_object(pDoc);
  Dwg_Object_BLOCK_HEADER* pMSpace = poMSpace->tio.object->tio.BLOCK_HEADER;
  dwg_point_3d start = { 0 };
  dwg_point_3d end = { 0 };
  for(size_t i = 0; i < ....;i++)
  {
    end.x = 10 * i;
    end.y = 10 * i;
    dwg_add_LINE(pMSpace, &start, &end);
  }
  dwg_write_file(..., pDoc);
  dwg_free_Document(pDoc);

src/dwg.c
```diff
@@ -3192,3 +3205,230 @@ dwg_log_dataflags (const int _loglevel, const int maxlevel,
         HANDLER (OUTPUT, "  vert_alignment (128)\n");
     }
 }
+
+static int
+Dwg_Object_Ref_compare (const Dwg_Object_Ref *key, const Dwg_Object_Ref **ppR)
```
rurban, line 3210: static methods don't need the `Dwg_` prefix. `ref_cmp` is enough.

src/dwg.c
```diff
@@ -3192,3 +3205,230 @@ dwg_log_dataflags (const int _loglevel, const int maxlevel,
...
+ *           returns NULL
+ *           *ppBefore pointer to key's insert pos in array
+ *
+ *Exceptions:
+ *   Input parameters are validated. Refer to the validation section of the
+ *function.
+ *
+ *******************************************************************************/
+static void *
+bsearch_ex (const void *key, const void *base, size_t num, size_t width,
```
rurban, line 3253: bsearch is in C99 already. Is this needed for VS2019?

src/dwg.c
gongxy-sh: OK. Thanks.

src/dwg.c
```diff
@@ -3192,3 +3205,230 @@ dwg_log_dataflags (const int _loglevel, const int maxlevel,
...
+        {
+          // key < lo
+          *ppBefore = lo;
+        }
+    }
+  return NULL;
+}
+
+void
+dwg_ordered_ref_appended (Dwg_Data *restrict dwg, Dwg_Object_Ref *ref)
```
rurban, line 3328: maybe just dwg_ordered_object_refs_add(). It's either insert or append

src/dwg.c
gongxy-sh: I need bsearch_ex to return the insertion point when the key is not found.

src/dwg.c
gongxy-sh: OK.

src/encode.c
```diff
@@ -5766,6 +5766,7 @@ find_prev_entity (Dwg_Object *obj)
   Dwg_Data *dwg = obj->parent;
   if (obj->supertype != DWG_SUPERTYPE_ENTITY)
     return NULL;
+  /*
   for (BITCODE_BL i = obj->index - 1; i > 0; i--)
```
rurban, line 5770: They are already searched backwards. Are there really so many non-entities? well, there are at the end, but still curious about the speedup.

src/dwg.c
rurban: Oh my, well then

src/dwg_api.c
```diff
@@ -23063,6 +23083,11 @@ dwg_insert_entity (Dwg_Object_BLOCK_HEADER *restrict _owner,
         }
     }
   in_postprocess_handles (obj);
+  //for speed up find_prev_entity
+  if (DWG_SUPERTYPE_ENTITY == obj->fixedtype)
+    {
+      dwg->prev_entity = obj->index;
```
rurban, line 23089: => prev_entity_index please.

gongxy-sh: OK

src/encode.c
gongxy-sh: lsearch
100000 entities used 00:03:34
300000 entities used 00:33:53

bsearch
100000 entities Used 00:01:19
300000 entities used  00:14:53

bsearch+prev_entity

100000 entities used 00:00:02
300000 entities used 00:00:23

rurban:
No, no extra memory management isolation functions. 
For free() we do have FREE_IF, the rest is standardized and could be overridden by boem-gc/jemalloc/tcmalloc/mimalloc/... headers.

gongxy-sh:
The following scenario: dwg writer, app debug mode/dwg library release mode

Dwg_ Data* dwg = dwg_ new_ Document(...);
...
dwg_ free(dwg);
free(dwg); -- This will cause crash, and memory leakage if not released

gongxy-sh:
By memory management isolation, I mean that whoever applies for memory will release it. In other words, the malloc/calloc/realloc/free functions called by the application and the dwg library may not be the same set of functions.

PR #650, opened 2023-02-28
Fix pre r13 tables

michal-josef-spacek:


rurban:
Nice msvc warnings in the new check!

PR #651, opened 2023-03-01
Optimization find_prev_entity,Accelerate DWG generation.

gongxy-sh:
About 50% faster DWG generation


rurban:
Nice. Small enough to be under the FSF CLA req's. Thanks a lot

rurban (approved):
LGTM and easy enough

PR #653, opened 2023-03-02
Optimize dwg_add_Handleref, at least an order of magnitude increase in DWG generation speed

gongxy-sh:
vs2019 test code
```
#include <windows.h>
#include <stdio.h>
#include <dwg_api.h>

int main(int argc, char *argv[])
{
  if (argc < 2)
  {
    printf("Usage %s num_objects [out_dwg]\n", argv[0]);
    return -1;
  }
  DWORD dwStart = GetTickCount();
  size_t num_objects = atoi(argv[1]);
  Dwg_Data* pDoc = dwg_new_Document(R_2000, 0, 0);
  Dwg_Object* poMSpace = dwg_model_space_object(pDoc);
  Dwg_Object_BLOCK_HEADER* pMSpace = poMSpace->tio.object->tio.BLOCK_HEADER;
  dwg_point_3d start = { 0 };
  dwg_point_3d end = { 0 };
  end.x = 10 * num_objects;
  for(size_t i = 0; i < num_objects;i++)
  {    
    end.y = 10 * i;
    dwg_add_LINE(pMSpace, &start, &end);
  }
  dwg_write_file(argc > 2 ? argv[2] : "dwgspeed.dwg", pDoc);
  dwg_free(pDoc);
  //free(pDoc);
  DWORD dwMS = (GetTickCount() - dwStart) / 1000;
  printf("%zu entities used %02d:%02d:%02d.\n", num_objects,
          dwMS / 3600, (dwMS % 3600) / 60, (dwMS % 3600) % 60);
  return 0;
}
```


gongxy-sh:
What can I do next to advance the progress ?

rurban:
See CONTRIBUTING

> See https://www.gnu.org/prep/maintain/html_node/Legally-Significant.html
and https://www.gnu.org/prep/maintain/html_node/Copyright-Papers.html
for details and contact us to obtain the relevant forms.
The most likely proper form is
http://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=blob_plain;f=doc/Copyright/request-assign.future;hb=HEAD

gongxy-sh:
Thanks.
I have sent an email according to your document the day before yesterday, but I haven't received a reply so far. I don't know if there is a mistake

rurban:
Ah great. They'll need at least a week or two

rurban:
In the meantime you can rebase it onto master

PR #659, opened 2023-03-13
Fix incorrect read w-coordinate for spline

esobolev321:
[NL-11_2000.zip](https://github.com/LibreDWG/libredwg/files/10956676/NL-11_2000.zip)
The file in which the spline was previously read incorrectly


rurban:
Thanks, trivial enough

PR #665, opened 2023-03-17
Fix TILEMODE header variables in preR13

michal-josef-spacek:


rurban:
RS really for a boolean?
Several more fixes for the removal needed. make check helps

michal-josef-spacek:
@rurban I think yes. Seems three RS variables in a row. I could set a bigger value for PLINECHECK and it's in DXF from AutoCAD.

I'm trying to figure out the value after PLINECHECK. It's in model space 0/1, but don't know what it's. If the number is bigger, it's cleared to "01 00" by AutoCAD.

The alternative version could be theoretically RC RC RS RC RC, but there are more variables than a number of variables in the header. I am for RS RS RS now.

rurban (changes requested):
And now just as in the next line. Cast the RS to B

rurban:
Merged, slightly fixed. Thanks!

PR #667, opened 2023-03-20
Fix last entity for block

esobolev321:


rurban (commented):
This looks tricky. I'll have to check later

esobolev321:
> This looks tricky. I'll have to check later

find_prev_entity returns the previous entity from the general list of objects, while entities owned by blocks are not part of this list

rurban:
I see, working on it in the `smoke/last_entity_for_block-pr667` branch

PR #670, opened 2023-03-21
Identify JUMP entity

michal-josef-spacek:
There were a mix of two polyline entities. Polyline is entity with id 19.
Entity with id 18 is jump to extra section and back.


rurban:
well, it's reference then. But ok with JUMP
I'll fix the missing unit-tests fo0r it

rurban:
Now it would be ready. stay with JUMP, or PLINEREF?

michal-josef-spacek:
@rurban I think that concept is common (for me is JUMP optimal).

PR #680, opened 2023-03-30
Add logging of 3dface invisible flag

michal-josef-spacek:


rurban:
I've renamed them to use the _3DFACE prefix and EDGE1-4

michal-josef-spacek:
@rurban ok, thanks

PR #699, opened 2023-04-21
Fix blocks_size for preR13

michal-josef-spacek:
Constant 0x40000000 was added in AC1001 version to blocks_size


rurban:
Please try a rebase and push --force. I just fixed the github actions

michal-josef-spacek:
@rurban I rebased

PR #711, opened 2023-04-26
Fix polyline 3d in preR13

michal-josef-spacek:


michal-josef-spacek:
This is bad guess of right polyline.

PR #740, opened 2023-05-25
Fix comments in VERTEX/POLYLINE entities

michal-josef-spacek:


rurban:
Well, I would rather keep/fix the R11 types for them

michal-josef-spacek:
This is only small cleanup of comments, still independent of others.
I have a full rewrite to preR13 POLYLINE/VERTEX. Waiting for the previous PR.

rurban:
No, I'd rather want to keep these types there, even if we get special entities for these

PR #742, opened 2023-05-30
Improve tests

michal-josef-spacek:


test/unit-testing/bits_test.c
```diff
@@ -561,6 +561,34 @@ bit_TV_to_utf8_tests (void)
...
+    fail ("bit_read_TF");
+
+  bitfree (&bitchain);
+}
+
+static void
+bit_write_TF_tests (void)
+{
+  Bit_Chain bitchain;
+  bit_write_TF (&bitchain, (BITCODE_TF)"GNU", 3);
```
rurban, line 637: overflow, this needs a `bitprepare (&bitchain, 3);`

rurban:
> @rurban I am a bit confused with this line: https://github.com/LibreDWG/libredwg/blob/master/src/bits.c#L2041 bit_write_TV()
> 
> ```
> ...
> if (dat->from_version < R_13b1)
>     bit_write_RS (dat, (BITCODE_RS)length);
> ...
> ```
> 
> There is dat->from_version and later dat->version. It's not bad?

yes, that looks wrong

rurban (approved):

PR #778, opened 2023-07-21
Fix LTYPE.strings_area conversion from JSON to DWG

michal-josef-spacek:


src/in_json.c
```diff
@@ -2192,7 +2192,8 @@ _set_struct_field (Bit_Chain *restrict dat, const Dwg_Object *restrict obj,
                   const size_t blen = len / 2;
                   // binary hexstring
                   BITCODE_TF buf = (BITCODE_TF)malloc (blen);
-                  if ((len = in_hex2bin (buf, str, blen) != blen))
+                  len = in_hex2bin (buf, str, blen);
+                  if ((len != blen))
```
rurban, line 2196: No (( please

rurban (changes requested):

src/in_json.c
michal-josef-spacek: In case of ```len = in_hex2bin (buf, str, blen) != blen``` len is 0 or 1 depending on result of ```in_hex2bin (buf, str, blen) != blen```.
I don't think that we need it.

if we have:
```
const size_t k = dwg->header.from_version > R_2004 ? 512 : 256;
```
and
```
if (len > k)
...
else if (len != k)
```

Or am I missing something?

rurban: Just the double parens

michal-josef-spacek: uff

rurban:
Merged

PR #782, opened 2023-07-21
Add tests for dwg_find_color_index()

michal-josef-spacek:


michal-josef-spacek:
The failure in mingw:
```
D:/a/_temp/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/13.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../../src/.libs/libredwg.dll.a(libredwg_0_dll_d000226.o):(.text+0x0): multiple definition of `dwg_find_color_index'; ../../src/.libs/dwg.o:D:/a/libredwg/libredwg/src/dwg.c:3013: first defined here
```
I don't know how to test and fix

rurban:
It's already exported. No need to link it again

michal-josef-spacek:
@rurban btw: in rgb_palette[256] is { 0xFF, 0xFF, 0xFF } two times (index 7 and 255).

rurban:
> @rurban btw: in rgb_palette[256] is { 0xFF, 0xFF, 0xFF } two times (index 7 and 255).

Yes, on purpose. It's the default color

rurban:
Merged with fixups

PR #791, opened 2023-07-31
Move address and offset to LOG_HANDLE

michal-josef-spacek:
This is better to diff between dwgread log outputs


rurban:
Yes, much better

PR #813, opened 2023-08-23
Fix summaryinfo conversions

michal-josef-spacek:


rurban:
Nope. As far as I remember this is now single byte < r2007, and double above

rurban (changes requested):
It's a T16 type, which means it must handle single < r12007 and double-bytes.

michal-josef-spacek:
Thanks for fix :-)

PR #821, opened 2023-08-28
error: ‘__builtin_strncat’ output truncated before terminating nul co…

jarkko-hautakorpi:
…pying as many bytes from a string as its length [-Werror=stringop-truncation]

header_variables_r11.spec:164:20: note: length computed here
164 |       size_t len = strlen ((char*)&_obj->MENUEXT[1]);

### DIFF
--- a/src/header_variables_r11.spec
+++ b/src/header_variables_r11.spec
@@ -161,7 +161,7 @@
 #endif
   DECODER {    
     if (_obj->MENUEXT[1]) {
-      size_t len = strlen ((char*)&_obj->MENUEXT[1]);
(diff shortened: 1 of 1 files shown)


rurban:
Good catch. Which compiler version detected that?

rurban (approved):

jarkko-hautakorpi:
> Good catch. Which compiler version detected that?

I'v got `gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.1)`

rurban:
merged with commit message reformatting, thanks

PR #834, opened 2023-09-05
Remove skipping of creating of BLOCK_HEADER in preR13

michal-josef-spacek:


rurban (changes requested):
r1.4/entities fails JSON roundtrips with this.

rurban:
I think this can be closed now

PR #842, opened 2023-09-18
Fix leader hook line flag

gentilinigm:
Implements a calculation method aligned with AutoCAD's hook lines display rules, it remains unclear if the flag is actually stored somewhere.

Reference documentation:

* AutoCAD Official Documentation: [link](https://help.autodesk.com/view/ACDMAC/2018/ENU/index.html?guid=GUID-BC466DEE-ACD8-419A-B017-AB3065336AD7)
* AutoCAD ActiveX API Reference: [link](https://documentation.help/AutoCAD-ActiveX-AAG/WS1a9193826455f5ff1a32d8d10ebc6b7ccc-6b57.htm)

Hook lines appear if the last leader line segment is at an angle greater than 15 degrees from horizontal. 
If the leader has no annotations or a spline path (?), it is certain that hook lines are not displayed.

### DIFF
--- a/src/dwg.spec
+++ b/src/dwg.spec
@@ -3406,10 +3406,9 @@ DWG_ENTITY (LEADER)
   FIELD_BD (box_width, 0);
   FIELD_B (hookline_dir, 0); // if hook line is on x direction
   FIELD_B (arrowhead_on, 0);
-  //ENCODER {
-  //  if (FIELD_VALUE (hookline_on))
-  //    FIELD_VALUE(arrowhead_type) |= 8;
-  //}
+  ENCODER {
+    HANDLE_HOOKLINE_FLAG(hookline_on, FIELD_VALUE(num_points), FIELD_VALUE(points), FIELD_VALUE(annot_type), FIELD_VALUE(path_type));
+  }
   FIELD_BSx (arrowhead_type, 0);
   DECODER {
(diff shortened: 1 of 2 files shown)


rurban (changes requested):
please squash them into 1 commit
didn't you add an angle helper (atan)?

gentilinigm:
>please squash them into 1 commit

Done!

> didn't you add an angle helper (atan)?

Yes, Ive used atan2: `angleR = atan2(pt1.y - pt2.y, pt1.x - pt2.x);`

rurban (changes requested):
We also need to calculate the hookline_on flag for the ENCODER, so better move the calculation to a helper function.

gentilinigm:
Sure I can move the calculation.

But how does the ENCODER work in this case exactly? Is it just a check?
I mean, the hook line status doesn't really tell us much about the other values.

rurban:
The problem is that we cannot always be sure that the user did set the hookline_flag correctly. We should set it according to the angle of the last 2 pts, same as in the DECODER.

gentilinigm:
I've made a few commits, isn't the hook line flag encoding DXF only though?

rurban:
I"ve rewritten it a bit, but kept your logic. Thanks!

PR #845, opened 2023-09-21
[fix] dead loop and segment fault bug

chunibyo-wly:
When I run dwgread on my own data, I faced a dead loop and segment fault.
I tried to fix it while I don't know if it is right.


rurban (approved):
looks very good, thanks a lot!
Can you just change the Subject line to "fix angle normalisations. arc_split" please, and force push

rurban:
Thanks, merged

PR #865, opened 2023-11-09
Fix 2ndheader encoding

michal-josef-spacek:


rurban:
Yes, it fixed a few json-check oda errors.

rurban:
We should store the SECTION_2NDHEADER_R13_SIZE part in some junk field

michal-josef-spacek:
I added info about structure to log.

rurban (changes requested):

include/dwg.h
```diff
@@ -9041,6 +9041,11 @@ typedef enum DWG_SECTION_TYPE_R13
   SECTION_THUMBNAIL_R13 = 6,
 } Dwg_Section_Type_r13;
 
+typedef enum DWG_SECTION_SIZE_R13
```
rurban, line 9044: I'd prefer this as define, not enum.

include/dwg.h
```diff
@@ -9633,6 +9638,7 @@ typedef struct _dwg_secondheader
   Dwg_SecondHeader_Handles handles[14];
   BITCODE_RS crc;
   BITCODE_RLL junk_r14; /*!< r14-r2000 */
+  BITCODE_RC before_sentinel[53];
```
rurban, line 9641: and use the define here

rurban (changes requested):

src/encode.c
```diff
@@ -3607,23 +3609,7 @@ dwg_encode (Dwg_Data *restrict dwg, Bit_Chain *restrict dat)
...
-          dwg->header.section[4].size = UINT64_C (4);
-          _obj->sections[4].size = UINT32_C (4);
-          dat->byte = pvzadr;
-          encode_secondheader_private (dat, dwg);
-        }
-
-      _obj->size = encode_patch_RLsize (dat, pvzadr);
-      if (_obj->size < dat->byte)
-        dwg->header.section[SECTION_2NDHEADER_R13].size = _obj->size;
+      encode_patch_RLsize (dat, pvzadr);
```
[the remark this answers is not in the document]
michal-josef-spacek: My code is totally wrong

include/dwg.h
michal-josef-spacek: My code is totally wrong

include/dwg.h
michal-josef-spacek: My code is totally wrong

rurban (approved):

src/encode.c
```diff
@@ -3527,10 +3518,6 @@ dwg_encode (Dwg_Data *restrict dwg, Bit_Chain *restrict dat)
       BITCODE_BL vcount;
 
       assert (dat->byte);
```
rurban, line 3520: add encode_objfreespace here

src/decode.c
```diff
@@ -916,7 +918,27 @@ decode_R13_R2000 (Bit_Chain *restrict dat, Dwg_Data *restrict dwg)
    */
 
   /*-------------------------------------------------------------------------
-   * Second header, section 3. R13c3-R2000 only.
+   * AcDb:ObjFreeSpace, section 3
+   */
+
+  if (dwg->header.sections > 3
+      && dwg->header.section[SECTION_OBJFREESPACE_R13].address != 0)
```
rurban, line 925: address is usually 0. rather search for the sentinel, and check if object_end equals dat->byte - 53. if there's room for this section before 2ndheader

rurban (changes requested):
see my work/r13-objfreespace

rurban:
Merged with fixes

Issue #871, opened 2023-11-24
Fix AuxHeader in AutoCAD 2000


michal-josef-spacek:
AuxHeader is broken in this time.
Example:
[BLANK_English.DWG.gz](https://github.com/LibreDWG/libredwg/files/13458155/BLANK_English.DWG.gz)
Output:
```
=======> AuxHeader:         97
         AuxHeader (end):  220
aux_intro: { 0xff 0x77 0x1 } [*RC]
dwg_version: 0x17 [RSx 0]
maint_version: 0x6 [RSx 0]
numsaves: 4294901760 [RL 0]
minus_1: 131071 [RLd 0]
numsaves_1: 0 [RS 0]
numsaves_2: 0 [RS 0]
zero: 1507328 [RL 0]
dwg_version_1: 0x6 [RSx 0]
maint_version_1: 23 [RS 0]
dwg_version_2: 0xa [RSx 0]
maint_version_2: 17134 [RS 0]
unknown_6rs: { 17134 0 1 0 0 0 } [*RS]
unknown_5rl: { 0 0 0 2310668288 954204197 } [*RL]
TDCREATE: [2310668980, 1313079333] 6321684-09-08 03:08:20 [TIMERLL 0]
TDUPDATE: [2884276, 0] 3184-10-11 00:00:00 [TIMERLL 0]
HANDSEED: 0x0 [RLx 0]
plot_stamp: 1 [RL 0]
zero_1: 0 [RS 0]
numsaves_3: 0 [RS 0]
zero_2: 0 [RL 0]
zero_3: 65536 [RL 0]
zero_4: 0 [RL 0]
numsaves_4: 0 [RL 0]
zero_5: 0 [RL 0]
zero_6: 0 [RL 0]
```

There is some move, because TDCREATE and TDUPDATE is bad.

Follow up for https://github.com/LibreDWG/libredwg/issues/860

michal-josef-spacek:
After #874:
```
=======> AuxHeader:         97
         AuxHeader (end):  220
aux_intro: { 0xff 0x77 0x1 } [*RC]
dwg_version: 0x17 [RSx 0]
maint_version: 0x6 [RSx 0]
numsaves: 1 [RL 0]
minus_1: -1 [RLd 0]
numsaves_1: 1 [RS 0]
numsaves_2: 0 [RS 0]
zero: 0 [RL 0]
dwg_version_1: 0x17 [RSx 0]
maint_version_1: 0x6 [RSx 0]
[...]

PR #874, opened 2023-11-25
Fix aux header

michal-josef-spacek:
Resolves: #871

### DIFF
--- a/src/auxheader.spec
+++ b/src/auxheader.spec
@@ -44,7 +44,7 @@
   FIELD_RSx (dwg_version, 0);
 #ifndef IS_JSON
   UNTIL (R_2013) {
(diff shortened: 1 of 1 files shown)


src/auxheader.spec
```diff
@@ -44,7 +44,7 @@
   FIELD_RSx (dwg_version, 0);
 #ifndef IS_JSON
   UNTIL (R_2013) {
-    FIELD_CAST (maint_version, RL, RSx, 0);
+    FIELD_RSx (maint_version, 0);
```
rurban, line 47: nope, maint_version is a RL. we need to cast it

rurban (changes requested):

PR #877, opened 2023-11-26
geomesh pts counted double

fr-an-k:
GEOMESH pts are a pair of coordinates, but the count was incorrectly increased for each, i.e. counting double.
It caused an invalid DWG error without log message.
This fixes it.


rurban:
Why did you close it? dxfread is not needed, and the geomesh fix looks good. Just force push without the 2nd commit

PR #886, opened 2023-12-02
no JSON whitespace in release version

fr-an-k:
Saves almost half the size; I plan to use this in a javascript production environment.


rurban (commented):
good idea, but better as runtime cmdline switch. like --json-compact or so

rurban:
I've used fmt=minJSON or outfile=xxx.min.json for this feature

rurban:
see work/json-ws-gh887-1

PR #887, opened 2023-12-02
No JSON whitespace in release version

fr-an-k:
See #886 - (I didn't manage to rebase the original feature branch that I accidentally branched of another feature branch in reasonable time).

I considered this feature as a runtime option, but it would be more work and I don't know how to compile the c file twice to produce the same function with a different flag, without creating a mess.

I could however, make this as an optional DISABLE_JSON_WHITESPACE flag rather than default when DEBUG is not set.

IMHO a sleak release version by default is a more important use case, and therefore should not include unnecessary whitespace or conditionals.

JSON readability is for debugging and should be done by piping, autoformat, setting the flag, 2 variants of each program, or using the debug version.

YAML/JSON5 would more suitable for readability anyway, and could also be done with macro's in out_json, but someone else can do that.

### DIFF
--- a/src/2ndheader.spec
+++ b/src/2ndheader.spec
@@ -24,7 +24,7 @@ VERSIONS (R_13, R_2000) {
   FIELD_TFF (version, 6, 0);
 #ifdef IS_JSON
   KEY (zero_5);
-  fprintf (dat->fh, "[ %d, %d, %d, %d, %d ]",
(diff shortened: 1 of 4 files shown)


fr-an-k:
Wait, I forgot to check the JSON, there's a small issue.

rurban (commented):
thanks, good work. bhut this needs a CLA from the FSF, which needfss some time

rurban:
I'll be doing it at runtime, by using a minJson format

PR #888, opened 2023-12-03
dxfread (and DISABLED_DXF/WRITE source inclusion issues)

fr-an-k:
This serves my use case of a javascript production environment (I'm using it already); converting to/from JSON uing CLI, until I can produce an npm or WASM equivalent of DXF <-> JSON converter.

Javascript is the most popular language dominating many platforms and it seems very uncommon for a library to not have bindings for it.

Converting indirectly through DWG is undesirable since this is very error prone and makes end users have to wait longer; the DXF files that I have to support are tens of MB.

Having this in the master branch would greatly simplify further development, especially since most of the DXF that I have to support are currently failing.


fr-an-k:
@rurban dxf_read_file was defined only when USE_WRITE was enabled.
I removed that part of the conditional to pass the tests, or was there a reason for that?
In that case I will solve it differently and add a comment to clarify.

fr-an-k:
The JSON issue was not related to this PR.
The JSON error appears to have been caused by an invalid entity.
In my code I skip failed entities in order to not abort the JSON, but I still need to free up and remove that half-finished entity.
I still need to double check the JSON of a DXF that actually works though.

fr-an-k:
Ok thanks, shoulnd't USE_WRITE then ideally only be used to include preR13 support in dxf_read_file?
Also, it is correct that all the programs should be turned off for --disable-write?
Otherwise they would pretty much be an empty executable; even dwglayers and dwggrep have write dependencies.

programs/dxfread.1
```diff
@@ -0,0 +1,42 @@
+.\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.48.5.
+.TH DXFREAD "1" "November 2023" "dxfread 0.13.0" "User Commands"
+.SH NAME
+dxfread \- manual page for dxfread 0.13.0
+.SH SYNOPSIS
+.B dxfread
+[\fI\,OPTION\/\fR]... \fI\,DXFFILE\/\fR
+.SH DESCRIPTION
+Reads the DWG into some optional output format to stdout or some file,
```
michal-josef-spacek, line 9: I think ```Reads the DXF```

rurban:
Sure, as in llvmfuzz

PR #889, opened 2023-12-03
Ignore/skip invalid objects

fr-an-k:
I have to handle many DXF that just gave error 0x800.
This PR adds a warning with line number if a DXF object/entity fails to parse.
It skips the rest of the object/entity and continues parsing rather than terminating.
Often the culprit is some exotic structure that I don't care about.
This PR gives me a way of logging (many cloud uploads) what went wrong, judge the severity and improve the code so it does parse.
It now does this by default, because IMHO that's what is expected, but a compile flag can be set to revert ot the old behavior (with improved warning).
The encode and out_** files ignore the invalid objects/entities as well, so a result can be produced.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ project(libredwg C)
 # Supported options -DLIBREDWG_LIBONLY=On
 #                   -DLIBREDWG_DISABLE_WRITE=On
 #                   -DLIBREDWG_DISABLE_JSON=On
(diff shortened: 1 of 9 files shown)


rurban:
I think this feature should be selectable at run-time, via some cmdline flag. Not compile-time.

rurban:
I"ll redo that for 0.13.2, with a run-time flag.

PR #896, opened 2023-12-15
Remove zero_5 variable

michal-josef-spacek:
This is about simplification of the header (and second header). It's for all DWG versions.

### DIFF
--- a/include/dwg.h
+++ b/include/dwg.h
@@ -9622,8 +9622,7 @@ typedef struct _dwg_secondheader
 {
   BITCODE_RL size;
   BITCODE_RL address;
-  BITCODE_RC version[7];
-  BITCODE_RC zero_5[5];
+  BITCODE_RC version[11];
   BITCODE_RC is_maint;
   BITCODE_RC zero_one_or_three;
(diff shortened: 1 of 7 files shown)


michal-josef-spacek:
I tested on all DWGs which I have, I tested JSON and DWG.

rurban:
Well, it removes the information that the version is of max length 8. but on the other side, it is most likely that the real code is more like that. So good enough

michal-josef-spacek:
There is probably a need for regen and doc.  

About JSONs: This is an interesting thing because JSON is not backward compatible in any way in the development process. I know there are many issues with structures (e.g. aux header is very problematic, and I don't see a way to fix still). This means there will be changes in JSON in the future.

michal-josef-spacek:
btw: I am planning some changes:
* I want to change is_maint to something better because this is "the maintenance version". The word "is" isn't right.
* I want to change dwg_version and maint_version, it is the version of AutoCAD in real, so something saved_by_autocad should be better. Seems that's one variable because of the BS in the Second Header. Probably better is to use the structure. Probably related to understand aux header.

PR #913, opened 2024-01-01
Fix OSMODE variable in header

michal-josef-spacek:
it's BS not BL


rurban:
And in the header? And in DXF?

michal-josef-spacek:
@rurban Your change seems ok

PR #928, opened 2024-01-29
Fix second header

michal-josef-spacek:


michal-josef-spacek:
@rurban I was looking to your actual changed to second header, but I don't think that's right.

michal-josef-spacek:
My actual proposal is to merge this.
The next step is to split objfreespace_second to objfreespace and second_header (I am working on PR, but with some weird error) This fixs the error with creating of second header without objfreespace.

rurban:
Yes, I also just found out about the section numbers and wrong addresses

Issue #860, opened 2023-11-03
Fix second header block


michal-josef-spacek:
This is a tracking issue for activity on the second header block.
I found issues in the case of https://github.com/LibreDWG/libredwg/issues/346

michal-josef-spacek:
Examples (all AC1012 DWG files):
* [BLANK.DWG.gz](https://github.com/LibreDWG/libredwg/files/13248284/BLANK.DWG.gz)
* [DEMO.DWG.gz](https://github.com/LibreDWG/libredwg/files/13248285/DEMO.DWG.gz)
* [LINE1.DWG.gz](https://github.com/LibreDWG/libredwg/files/13248286/LINE1.DWG.gz)

BLANK.DWG:
* Blank drawing saved in AutoCAD R13
Actual header
```
This file's version code is: AC1012 (r13)
zero_5: { 0x0 0x0 0x0 0x0 0x0 } [*RC]
is_maint: 0x0 [RC 0]
zero_one_or_three: 0x1 [RC 0]
thumbnail_address: 562 [RL 0]
dwg_version: 0x0 [RC 0]
maint_version: 0x0 [RC 0]
codepage: 12 [RS 0]
CP850

num_sections: 3 [RL]
num_sections => 3
section[0].number:     0 [RC] AcDb:Header
section[0].address:   70 [RL]
section[0].size:     454 [RL]
section[1].number:     1 [RC] AcDb:Classes
section[1].address:  524 [RL]
section[1].size:      38 [RL]
section[2].number:     2 [RC] AcDb:Handles
section[2].address: 3977 [RL]
section[2].size:      79 [RL]
crc: A24C [RSx] from 0-52
         HEADER (end):      70
```
Actual Second header
```
=======> Second Header 3 (start):     4056
size: 90 [RL 0]
address: 4056 [BL 0]
version: "AC1012" [TFF 11 0]
4143313031320000000000
0xf [RC 0]
null_b: { 0 0 0 0 } [*B]
unknown_10: 0x18 [RC 0]
unknown_rc4: { 0x30 0x1 } [*RC]
num_sections: 3 [RC 0]
sections[0].nr: 0 [RCd 0]
sections[0].address: 70 [BL 0]
sections[0].size: 454 [BL 0]
sections[1].nr: 1 [RCd 0]
sections[1].address: 524 [BL 0]
sections[1].size: 38 [BL 0]
sections[2].nr: 2 [RCd 0]
[...]

PR #929, opened 2024-01-29
WIP Split objfreespace and second header

michal-josef-spacek:
There is possibility of second header and without objfreespace in AC1012. This mean, that we need to split.

### DIFF
--- a/include/dwg.h
+++ b/include/dwg.h
@@ -9054,10 +9054,11 @@ typedef enum DWG_SECTION_TYPE_R13
   SECTION_HEADER_R13 = 0,
   SECTION_CLASSES_R13 = 1,
   SECTION_HANDLES_R13 = 2,
-  SECTION_OBJFREESPACE_R13 = 3, /* including the 2ndheader */
+  SECTION_OBJFREESPACE_R13 = 3,
   SECTION_TEMPLATE_R13 = 4,
   SECTION_AUXHEADER_R2000 = 5,
   SECTION_THUMBNAIL_R13 = 6,
+  SECTION_SECOND_HEADER_R13 = 7,
 } Dwg_Section_Type_r13;
 
 typedef enum DWG_SECTION_TYPE_R11 /* tables */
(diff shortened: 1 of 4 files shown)


michal-josef-spacek:
@rurban First version, with error, which I don't see
```
In file included from encode.c:53:
encode.c: In function ‘encode_template’:
common.h:446:6: error: this condition has identical branches [-Werror=duplicated-branches]
  446 |   if (dat->version >= v)
      |      ^
encode.c:2506:3: note: in expansion of macro ‘SINCE’
 2506 |   SINCE (R_2004)
      |   ^~~~~
```

michal-josef-spacek:
Ha, another case.

AC1012 - some versions do have not objfreespace record in the header, the section doesn't exist.
other AC1012/AC1013/AC1014/AC1015 < AutoCAD 2005 - has objfreespace record in the header, section exists
AC1015 > AutoCAD 2005 - has objfreespace record in the header with 0 size, section doesn't exist.

rurban:
section 3 often does not exist, i.e. empty address and size, but often the 2ndheader is still attached to it. My latest fixes work now. I don't want to split it, as it confuses the section number logic. (dwg->header.sections)

include/dwg.h
```diff
@@ -9037,10 +9037,11 @@ typedef enum DWG_SECTION_TYPE_R13
   SECTION_HEADER_R13 = 0,
   SECTION_CLASSES_R13 = 1,
   SECTION_HANDLES_R13 = 2,
-  SECTION_OBJFREESPACE_R13 = 3, /* including the 2ndheader */
-  SECTION_TEMPLATE_R13 = 4,
-  SECTION_AUXHEADER_R2000 = 5,
-  SECTION_THUMBNAIL_R13 = 6,
+  SECTION_OBJFREESPACE_R13 = 3,
+  SECTION_SECOND_HEADER_R13 = 4,
```
rurban, line 9041: Nope this confuses our section numbering even more. I'd like it keeping attached to the ObjFreeSpace

rurban (changes requested):

include/dwg.h
michal-josef-spacek: I need to look again.

michal-josef-spacek:
@rurban I updated PR.
See comment https://github.com/LibreDWG/libredwg/issues/860#issuecomment-1918511755

rurban:
For the time being I did it simplier fir 0.13, as your patch with my fixes filled up my hard disc.

michal-josef-spacek:
Hm, there is some memory issue.

rurban:
But we should fix this better

PR #932, opened 2024-02-04
Fix PLINEGEN flag in LWPOLYLINE

michal-josef-spacek:
In DWG file is PLINEGEN defined by 256, in DXF by 128.


michal-josef-spacek:
Tested on dwgread, conversion from dwg to dxf and back.

PR #933, opened 2024-02-04
Fix comments of ObjFreeSpace and SecondHeader

michal-josef-spacek:


michal-josef-spacek:
@rurban Could we merge this?

rurban (approved):

PR #936, opened 2024-02-08
Fix summaryinfo props

michal-josef-spacek:


rurban:
Much better, thanks, but not much coverage. I"ll regen-dynapi and merge it

rurban:
Merged

Issue #943, opened 2024-02-27
Windows long and signed JSON conversion


michal-josef-spacek:
The conversion from JSON to DWG is not same on Windows as on Linux.

Examples on libredwg.0.13.3:
[linux.tar.gz](https://github.com/LibreDWG/libredwg/files/14424125/linux.tar.gz)
[wine.tar.gz](https://github.com/LibreDWG/libredwg/files/14424126/wine.tar.gz)

Diff between dwgread output of Windows dwgrewrite and conversion versions:
```
1c1
< Reading DWG file line00-rewrite.dwg
---
> Reading DWG file out00.dwg
19,20c19,20
< section[2].address:    33136 [RL]
< section[2].size:         496 [RL]
---
> section[2].address:    32998 [RL]
> section[2].size:         498 [RL]
25c25
< section[4].address:    33780 [RL]
---
> section[4].address:    33644 [RL]
30c30
< crc: 1D68 [RSx] from 0-79
---
> crc: CA1D [RSx] from 0-79
272,273c272,273
< DIMLWD: -2 [BSd 70]
< DIMLWE: -2 [BSd 70]
---
> DIMLWD: -1 [BSd 70]
> DIMLWE: -1 [BSd 70]
316c316
< crc: 38FA [RSx] from 24955-25438=483
---
> crc: 68C1 [RSx] from 24955-25438=483
341,344c341,344
< =======> Handles (start) :    33136
<          Handles (end)   :    33632
<          Length: 496
< Handles page size: 490 [RS_BE]
---
> =======> Handles (start) :    32998
>          Handles (end)   :    33496
>          Length: 498
> Handles page size: 492 [RS_BE]
471c471
< Object number: 7/7, Size: 15 [MS], Type: 66 [BS]
---
> Object number: 7/7, Size: 18 [MS], Type: 66 [BS]
474c474
<  Hdlsize: 0x30,
---
>  Hdlsize: 0x48,
479c479
< num_entries: 2 [BS 70]
---
> num_entries: 3 [BS 70]
484,485c484,486
< crc: 5E17 [RSx]
<  check_CRC 17: 5E17 == 5E17
---
> entries[2]: (2.2.12D) abs:301 [H* 0]
> crc: A957 [RSx]
>  check_CRC 20: A957 == A957
801,802c802,803
< crc: 9910 [RSx]
<  check_CRC 65: 9910 == 9910
---
> crc: 7CBD [RSx]
[...]

PR #952, opened 2024-03-09
Add missing macros for DWG to JSON conversions

michal-josef-spacek:


michal-josef-spacek:
Fixes two cases in https://github.com/LibreDWG/libredwg/issues/943

rurban (changes requested):
0x hex numbers are invalid in json.

rurban:
I fixed it better

michal-josef-spacek:
Thank you

PR #957, opened 2024-03-12
Improve logging of text dashes

michal-josef-spacek:


michal-josef-spacek:
I think, that we could improve this. To something ```dashes[1].text: FOO [TF 9]```

michal-josef-spacek:
Actual logging output:
```
...
strings_area: "FOO" [TF 256 0]
dashes[1].text: "FOO" [TF 3 9] @0
dashes[4].text: "BAR" [TF 3 9] @4
...
```

rurban:
I'd rather like 
strings_area: "FOO\000BAR\000BAZ"

michal-josef-spacek:
> I'd rather like
> strings_area: "FOO\000BAR\000BAZ"

no problem, but why not "FOO\0BAR" ?

rurban:
> > I'd rather like
> > strings_area: "FOO\000BAR\000BAZ"
> 
> no problem, but why not "FOO\0BAR\0BAZ" ?

This would look nicer, but is invalid C. Only \000 or \x00 would do

michal-josef-spacek:
We are talking about logging, this is for "user" :-) I don't understand C validity in this case.

rurban:
Well, then go for it. Looks better

rurban:
I"ve just fixed the asan issue with 17f95b23d07895ed289fbf1f6acbd4c792ace816, please rebase on latest master soon.

michal-josef-spacek:
Done

PR #980, opened 2024-05-31
Add test data for R2000

michal-josef-spacek:
The *scr files are the 2d/3d entity sets and the DWG/DXF files are the script command output in R2000.


rurban:
They need to be added to Makefile.am DATA also

michal-josef-spacek:
I added files to Makefile.am

PR #1087, opened 2025-02-28
Add ".0" to BD, RD fields if not present

michal-josef-spacek:
The result is similar as from AutoCAD.
It helps with naive diff between AutoCAD outputs and DXF outputs from LibreDWG.

### DIFF
--- a/src/out_dxf.c
+++ b/src/out_dxf.c
@@ -29,6 +29,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


rurban:
The strings "NAN" and "E" are highly libc and format arg dependent

rurban:
I think best is to check all chars for isdigit, which implies no comma, nan, exp, inf

rurban (changes requested):
Too fragile, see my comments for something simplier

PR #1181, opened 2025-09-05
Multi-platform CI expansion with downloadable artifacts

neilberkman:
## Multi-platform CI expansion with downloadable artifacts

This PR expands the GitHub Actions CI to build, test, and generate downloadable artifacts across multiple platforms.

### What's new
- Linux: Ubuntu (20.04, 22.04, latest), Fedora, Debian, Alpine, Rocky Linux
- macOS: 12, 13, latest with gcc and clang
- Windows: MinGW and MSVC builds
- Big-endian: PowerPC64, MIPS32
- **Downloadable artifacts for all platforms** (tar.gz for Linux/macOS, zip for Windows)

### Coverage
- ~90% of Linux distro market share
- All major operating systems
- Both endianness architectures
- Multiple compilers (gcc, clang, MSVC)

### Artifacts
Each successful build produces downloadable artifacts that users can get directly from GitHub Actions runs without needing to compile.

### Testing
Tested on fork: https://github.com/neilberkman/libredwg/actions/runs/17504170044

All platforms pass except Fedora which has one test failure that appears to be a pre-existing issue now caught by Fedora's stricter compilation flags.


neilberkman:
@rurban this should be clean now. You can see a sample release from the most recent build here: https://github.com/neilberkman/libredwg/releases/tag/0.13.3.7835.5_b068c.7492

PR #1192, opened 2025-10-31
Amend CLOSE flag

danikolovv:
Value of o->flag must be at least 512.


rurban (changes requested):
Nope. Convert 1 to 512

danikolovv:
Hello, 

> Your patch overwrites all other DXF flags, and just sets CLOSED

Yes, I got acquainted with this and for that reason reverted the changes to the state before this pull request.

> looks correct to me. There must be some other problem, I think

Yes, for that reason as a maintainer of this project can you point out, in which file is flags reading of a DWG file implemented?

Issue #1187, opened 2025-10-21
Mechanical drawing symbol missing


ywu2014:
i'm using dwg2dxf commond to convert dwg file to dxf file.my dwg file has some mechanical drawing symbols, for examples section 2 in figure below is degree of concentricity
<img width="559" height="128" alt="Image" src="https://github.com/user-attachments/assets/c6c9deca-40d3-47d8-82f7-646dd9c8f22b" />. 
after convert, i did get the dxf file, but when i parse the dxf file using ezdxf in python, i can read content in section 1, content is \A1; φ1{\H0.7x;\S+0^-0.05;}, but i can't read content in section 2.
so i open the dxf file direct using text editor，and i can find section 1 content，show as below：

<img width="667" height="859" alt="Image" src="https://github.com/user-attachments/assets/4f6714a3-8aa2-4d68-b7e6-31361cfe91cd" />

but i can't find section 2 content in dxf file. i don't know the reason. is the section 2 content missing when convert from dwg file to dxf file? if so,since my dwg file has many mechanical drawing symbol, how can i remain these symble after i convert dwg files to dxf file?

ywu2014:
@michal-josef-spacek thanks. i tried your TOLERANCE.DWG file, it works. but in my scene，my dwg file was created by relatively new version of CAD Editor(some of them i known exact version， some i don't known), for example:
[TOLERANCE_NEW_VERSION.dwg.gz](https://github.com/user-attachments/files/23110997/TOLERANCE_NEW_VERSION.dwg.gz)
my libredwg version is 0.13.3.7835 running on ubuntu 22.04, when i tried to convert, some warning and errors appear in console,like 
`
Reading DWG file TOLERANCE_NEW_VERSION.dwg
ERROR: Skip section AcDb:Preview with max decompression size 0x6da0 > 0x1800
Warning: Unstable Class object 502 MATERIAL (0x481) 59/0
Warning: Unstable Class object 502 MATERIAL (0x481) 60/0
Warning: Unstable Class object 502 MATERIAL (0x481) 61/0
Warning: Unstable Class object 504 TABLESTYLE (0xfff) 81/0
Warning: TODO TABLESTYLE r2010+ missing fields
Warning: Unstable Class object 505 MLEADERSTYLE (0xfff) 83/0
Warning: Unstable Class object 511 ACDB_TEXTOBJECTCONTEXTDATA_CLASS (0x481) 437/0
[...]

PR #1194, opened 2025-11-02
Fix r2007 TOLERANCE entity

michal-josef-spacek:
TOLERANCE entity need a has_strings bit.

Fixes https://github.com/LibreDWG/libredwg/issues/1187


rurban (approved):

rurban:
This was on the 42th hour of my flight back. But looks very good to me

PR #1210, opened 2026-01-01
Fix DXF encoding: convert TV strings to UTF-8

matabar:
## Summary

Fixes encoding issues where LibreDWG was outputting DXF files with:
- UTF-8 encoded bytes in text fields  
- But declaring CP1252 codepage in `$DWGCODEPAGE` header

This caused ezdxf to read the UTF-8 bytes as CP1252, creating surrogate characters (`\udcxx`) for undefined bytes in CP1252.

## Changes

1. **dwg_api.c**: Set default codepage to `CP_UTF8` (0) instead of `CP_ANSI_1252` (30)
2. **out_dxf.c**: Convert TV strings from source codepage to UTF-8 before output
3. **out_dxf.c**: Always declare UTF8 codepage in DXF output

## Impact

- DXF files will have properly converted UTF-8 text
- Compatible with standard DXF readers like ezdxf
- No more surrogate character encoding errors

### DIFF
--- a/src/dwg_api.c
+++ b/src/dwg_api.c
@@ -24910,7 +24910,7 @@ dwg_add_Document (Dwg_Data *restrict dwg, const int imperial)
   // dwg->header.dwg_version = 0x17; // prefer encode if dwg_version is 0
   dwg_ver_struct = (struct dwg_versions *)dwg_version_struct (version);
   if (!dwg->header.codepage)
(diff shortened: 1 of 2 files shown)


matabar:
Wrong repository - closing

rurban:
The default encoding depends on the version. With r2000 it's 30 (Latin-1), only with later versions it's UTF-8.

And for such large PR's we need FSF contribution assignments.

Issue #1216, opened 2026-01-18
JSON Output: Layer Color Index does not correlate to RGB value


MarcWeberFS:
When converting a DWG into a JSON file, theres an issue with the nested color index. When the index is nested inside of color, the index is 256 and the rgb value is defined as c3000073.

```json
"color": {
        "index": 256,
        "rgb": "c3000073"
 },
```

The expected value for index in the example is 115.

---

*Steps to reproduce:*

1. run ```dwgread.exe -O json 12345-01_A-SC-S01-050-303Rb-1.dwg > output.json```
2. Search for Layer with the name ```"name": "M5_DECKENBEKLEIDUNGEN"```
3. Look at the color object and its nested values.

[12345-01_A-SC-S01-050-303Rb-1.zip](https://github.com/user-attachments/files/24698511/12345-01_A-SC-S01-050-303Rb-1.zip)

PR #1217, opened 2026-01-18
1216 - fix: Layer Color Index now matches the RGB value

MarcWeberFS:
This is the fix for ticket [1216](https://github.com/LibreDWG/libredwg/issues/1216)

Technical Analysis: The RGB value is 0xc3000073.
- High Byte (0xc3): Indicates the color method is ByACI (AutoCAD Color Index).
- Low Byte (0x73): Hex 73 equals Decimal 115. It appears the parser is not extracting the low byte when the 0xc3 flag is present in the nested color object.

### DIFF
--- a/src/out_json.c
+++ b/src/out_json.c
@@ -612,12 +612,35 @@ field_cmc (Bit_Chain *dat, const char *restrict key,
     {
       KEYs (_path_field (key));
       HASH;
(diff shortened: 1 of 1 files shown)


rurban (commented):
Looks good at first sight, but lots of cases to test. Thanks!

PR #1225, opened 2026-03-24
Fix compilation issues on modern systems with GCC

guifavretto:
In systems with modern glibc (like Arch Linux), including system
headers before config.h could lead to _POSIX_C_SOURCE macro
redefinition errors. Added include guards to config.h and a #ifndef
protection for _POSIX_C_SOURCE to prevent this.
Also, fixed a -Werror=discarded-qualifiers warning when casting
the result of strstr to char * in bits.c.

* configure.ac: Add AH_TOP and AH_BOTTOM for CONFIG_H include guards.
  Use AH_VERBATIM to wrap _POSIX_C_SOURCE in #ifndef.
* src/bits.c (bit_is_U_expand, bit_is_M_expand): Add explicit (char *)
  cast to the strstr return value.
* src/config.h.in: Regenerated to reflect the new include guards and
  macro protections.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -50,6 +50,9 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
 dnl create a config.h file (Automake will add -DHAVE_CONFIG_H)
 AC_CONFIG_HEADERS([src/config.h])
+AH_TOP([#ifndef CONFIG_H
+#define CONFIG_H 1])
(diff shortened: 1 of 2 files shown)


rurban (changes requested):
Can you please seperate them into 2 commits. configure and bits. 
- configure looks good 
- bits looks good.

Both look trivial, no CLA to the FSF needed, thanks

codefactor seems to be broken, let's ignore it.

PR #1235, opened 2026-04-05
Fix DXF codes in ATTDEF entity

michal-josef-spacek:


rurban (changes requested):
not really. DXF {} has its own block before, with tag and prompt already. dont emit it twice.

Just the flags 70 could be mixed up with mtext_type 70. havent checked that yet

PR #1277, opened 2026-06-11
OSS-Fuzz: Expand coverage for llvmfuzz.c with output converter/version

arthurscchan:
This PR expands the coverage of llvmfuzz by deriving the converter and version options from the fuzz input, rather than leaving them at their default 0 setting. Previously the libFuzzer path only ever exercised dwg_encode; with this change the fuzzer also reaches the other converters across a range of version combinations, exercising the logic that depends on them and expand the fuzzing coverage and effectiveness.


rurban (approved):

rurban:
Thanks, merged manually

Issue #1278, opened 2026-06-12
R2018 MTEXT extents_width/extents_height appear to be decoded swapped


ZalivahaAM:
Hi. I found what looks like a LibreDWG decoding issue for R2018 DWG MTEXT entities.

For the same MTEXT saved as R2013 and R2018, the geometry and orientation fields are identical:
- x_axis_dir = (1, 0, 0)
- extrusion = (0, 0, 1)
- angle = 0
- insertion point is the same

But extents_width and extents_height are swapped in the R2018 DWG.

R2013:
extents_width = 19.765347885402456
extents_height = 6.5129604365620724

R2018:
extents_width = 6.5129604365620724
extents_height = 19.765347885402456

This makes horizontal MTEXT render as a vertical/narrow text box in applications using LibreDWG.

I also noticed this in src/dwg.spec for R2018 MTEXT embedded/redundant fields:

DXF branch:
FIELD_BD(extents_width, 42);
FIELD_BD(extents_height, 43);

binary DWG branch:
FIELD_BD(extents_height, 43);
FIELD_BD(extents_width, 42);

This looks reversed for binary R2018 MTEXT.

Could you please check if the binary R2018 MTEXT field order should be width first, height second?

rurban:
Thanks. Also found by someone, and already fixed.

PR #1279, opened 2026-06-14
Fix swapped MTEXT extents_width/extents_height for R2018 (#1278)

informatimago:
Fixes the R2018 (AC1032) MTEXT `extents_width`/`extents_height` swap reported in #1278.

## Problem

For R2018, a wide-and-short MTEXT is decoded with width and height swapped — e.g. a text that is ~124 wide × 26 tall comes out as `extents_width=26.14, extents_height=124.09`. AutoCAD reversed the on-disk extents order somewhere in (R2013 … R2018]; the DXF group codes (`42` = width ≤ `41`, `43` = height) are already correct — only the **binary field order** was wrong.

## Root cause

The visible swap comes from the R2018-only **embedded** MTEXT object (`AcDbMTextObjectEmbedded`), which read the two extents doubles **height-first**. The plain MTEXT extents (`src/dwg.spec` ~3343) already read correctly and are overwritten by the embedded "redundant fields" block, so the fix is in the embedded paths:

- `src/dwg.spec`: the `SINCE (R_2018)` `is_not_annotative` embedded block read the binary height-first (DXF was width-first) — now **width-first** for both.
- `src/dwg_spec_shared.h`: `AcDbMTextObjectEmbedded` (also used by R2018 multiline `ATTRIB`) — now **width-first**.

The `FIELD_BD`/`SUB_FIELD_BD` macros are bidirectional, so decode and encode are fixed together. Both paths are R2018-only, so R2013-and-earlier MTEXT is untouched and no version gate is needed.

## Verification

On an R2018 drawing with a wide MTEXT, `dwgread -O JSON` now reports `extents_width≈124.09 / extents_height≈26.14` (was swapped), with `extents_width ≤ rect_width`, and the plain and embedded extents now agree. A `dwgrewrite --as r2018` round-trip is stable.

### DIFF
--- a/src/dwg.spec
+++ b/src/dwg.spec
@@ -3385,13 +3385,11 @@ DWG_ENTITY (MTEXT)
         }
         FIELD_BD (rect_width, 40);
         FIELD_BD (rect_height, 41);
-        DXF {
-          FIELD_BD (extents_width, 42);
-          FIELD_BD (extents_height, 43);
-        } else {
-          FIELD_BD (extents_height, 43);
-          FIELD_BD (extents_width, 42);
-        }
+        // GH #1278: the R2018 embedded MTEXT stores the extents width-first
+        // (1st double = width), same as DXF. (Earlier the binary was read
+        // height-first here, swapping width<->height for wide-and-short text.)
+        FIELD_BD (extents_width, 42);
+        FIELD_BD (extents_height, 43);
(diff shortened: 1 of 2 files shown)


rurban (approved):

rurban:
Thanks, fixed by myself, attributing you of course

PR #1280, opened 2026-06-14
R2018 write/encode fidelity: object loss, thumbnail, DXF UTF-8/codepage, MTEXT, LAYER colour

informatimago:
A series of fidelity fixes for the modern (R2018 / AC1032) read↔write paths, found while round-tripping real R2018 drawings.

### Commits

1. **Fix swapped MTEXT extents for R2018 (#1278)** — embedded MTEXT extents read width-first. Also submitted standalone as #1279 (this branch is based on it).
2. **encode: fix R2018 decode→encode→decode object loss + thumbnail growth** — `UNKNOWN_OBJ`/`UNKNOWN_ENT` proxy blobs were dropped on a same-version DWG→DWG re-encode (the raw re-emit was gated on `DWG_OPTS_INJSON`; now also for DWG input), and the `AcDb:Preview` thumbnail grew 16 bytes per round-trip (a duplicate `THUMBNAIL_END` sentinel). `example_2018` now round-trips 474/474 objects with a byte-identical, stable thumbnail.
3. **in_dxf: concatenate MTEXT text split across group codes 3 and 1** — long MTEXT text (>250 chars) was overwritten chunk-by-chunk and truncated to one fragment on import; now concatenated.
4. **bits: decode UTF-8 DXF strings to UTF-16 when encoding to r2007+** — DXF (UTF-8) strings were zero-extended byte-wise when written to an r2007+ target, double-encoding non-ASCII (`été`→`Ã©tÃ©`); now decoded as real UTF-8, gated on `DWG_OPTS_INDXF` so the JSON/add-API codepage paths are unchanged.
5. **in_dxf: preserve leading spaces in string values; fix r2007+ DWGCODEPAGE read** — `dxf_read_string` stripped significant leading spaces (breaking MTEXT continuations), and `DWGCODEPAGE` (a `BITCODE_TV`, i.e. UTF-16 for r2007+) was read as a C string so `ANSI_1252` was seen as `A`.
6. **out_dxf: only negate LAYER colour (DXF 62) when the layer is off** — the exporter unconditionally negated group-62, so layers that are **on** were exported as **off** (their entities became invisible, and a DWG→DXF→DWG round-trip flipped them off); now negated only when `off`.
7. **build: don't redeclare memmem() on macOS** — small build fix (macOS declares `memmem()` in `<string.h>`, tripping `-Werror=redundant-decls`).

### Testing

Verified against real R2018 files and the test corpus via DWG→DWG (`dwgrewrite`), DXF (`dwg2dxf | dxf2dwg`), and JSON round-trips: no object loss, byte-identical MTEXT/accented text, stable thumbnail, correct LAYER on/off. Existing r13–r2000 paths are unaffected.

### Note on overlap

Commit 1 is the #1278 fix, also opened as the smaller standalone PR #1279; this branch includes it as its base so the encoder work builds on it. Once #1279 lands, this branch can be rebased to drop the duplicate.

### DIFF
--- a/src/bits.c
+++ b/src/bits.c
@@ -2282,25 +2282,63 @@ bit_write_T (Bit_Chain *restrict dat, BITCODE_T restrict s)
               const char *endp = s + len;
               BITCODE_TU ws = (BITCODE_TU)malloc ((len + 1) * 2);
               const BITCODE_TU orig = ws;
+              // DXF in-memory strings are UTF-8 (in_dxf keeps them as read);
+              // decode multi-byte sequences to UTF-16 so non-ASCII is not
+              // double-encoded by a bare byte copy. Other producers of TV here
+              // (JSON, add API, old-DWG up-convert) already hold codepage bytes
+              // with non-representable chars as \U+xxxx escapes, so they keep
+              // the byte-wise path. Both honour the inline \U+xxxx escape.
+              const bool is_utf8 = (dat->opts & DWG_OPTS_INDXF) != 0;
               while (s < endp)
                 {
-                  uint16_t c = *s++;
-                  // in this case the resulting len is shorter
-                  if (c == '\\' && s[0] == 'U' && s[1] == '+' && ishex (s[2])
-                      && ishex (s[3]) && ishex (s[4]) && ishex (s[5]))
+                  unsigned char c = (unsigned char)*s;
+                  // \U+xxxx escape: shorter result
+                  if (c == '\\' && s + 6 < endp + 1 && s[1] == 'U'
+                      && s[2] == '+' && ishex (s[3]) && ishex (s[4])
+                      && ishex (s[5]) && ishex (s[6]))
                     {
                       unsigned x;
-                      if (sscanf (&s[2], "%04X", &x) > 0)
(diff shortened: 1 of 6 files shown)


rurban:
Excited to get a fellow lisp hacker on board! Also excited to see clautolisp! 
Unfortunately that's too much code to add with a FSF CLA. If you add it as tickets chunk by chunk I'll add it by myself. A CLA needs about a month to go through.
The analysis is more important than the code anyway.

informatimago:
> Excited to get a fellow lisp hacker on board! Also excited to see clautolisp! Unfortunately that's too much code to add with a FSF CLA. If you add it as tickets chunk by chunk I'll add it by myself. A CLA needs about a month to go through. The analysis is more important than the code anyway.

Ok, so I would make say, one PR per commit?
Otherwise, I think I might contribute more, so let's start the CLA process.

informatimago:
Superseded: splitting this into one PR per commit for easier independent review and merge. Each fix is being resubmitted as its own focused PR (the MTEXT extents #1278 fix is already #1279). Closing this umbrella PR.

PR #1284, opened 2026-06-15
in_dxf: concatenate MTEXT text split across group codes 3 and 1

informatimago:
MTEXT text longer than 250 characters is emitted in DXF as one or more
250-char chunks under group code 3 followed by a final (<250) chunk under
group code 1. The importer overwrote o->text on each code 3 (and only
reallocated when the new chunk was longer), and let the final code 1 fall
through to the generic field setter, which overwrote again. The result was
a single surviving fragment — a 1497-char paragraph imported as one byte.

Handle codes 3 and 1 together for MTEXT and append each chunk (a short
MTEXT still has only the code 1, handled as the first chunk). On
dwg2dxf | dxf2dwg of a real R2018 multiline MTEXT the full text and all
\P paragraph breaks now survive; short single-chunk MTEXT is unchanged.

### DIFF
--- a/src/in_dxf.c
+++ b/src/in_dxf.c
@@ -11154,28 +11154,39 @@ static __nonnull ((1, 2, 3, 4)) Dxf_Pair *new_object (
                 LOG_WARN ("Unhandled LAYOUT.1 in subclass %s", subclass);
               goto next_pair;
             }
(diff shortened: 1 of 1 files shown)


rurban (approved):

rurban:
Tested ok

PR #1288, opened 2026-06-15
OSS-Fuzz: Add new fuzzer targets dwgapi

arthurscchan:
This PR adds a new OSS-Fuzz fuzzer targets dwg apis.


rurban:
Nice!. 
But this is too big for a non-CLA contribution. Check CONTRIBUTING and apply for a copyright assignment. Needs about a month or so.

arthurscchan:
Thanks for your prompt reply. I will handle and follow the CLA procedure.

PR #1290, opened 2026-06-18
out_dxf: only negate LAYER colour (DXF 62) when the layer is off

informatimago:
The LAYER table writer negated the group-62 colour index unconditionally, so
every ON layer was exported with a negative 62 (= "layer off") and its
entities became invisible after dwg2dxf; a DWG->DXF->DWG round-trip flipped
the layer off. libredwg's own decoder reports the layer as on, so this was an
exporter-only inconsistency.

Derive the sign from the decoded `off` flag: negate only when off, using
abs() so an already-signed index cannot double-negate, with explicit braces.
An ON layer now exports its true positive ACI (62=7 instead of -7) and an OFF
layer exports negative; the DWG->DXF->DWG round-trip preserves on/off.

This re-applies the fix from the earlier #1283 (which did not land on master)
and incorporates the review feedback there (sign via -abs(), braces).

### DIFF
--- a/src/dwg.spec
+++ b/src/dwg.spec
@@ -3817,8 +3817,15 @@ DWG_TABLE (LAYER)
       if (FIELD_VALUE (color.index) == 256) {
         FIELD_VALUE (color.index) = FIELD_VALUE (color.rgb) & 0xFF;
       }
-      // Negative value in case of disabled layer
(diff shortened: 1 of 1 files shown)


rurban (approved):
I thought I used abs already

PR #1291, opened 2026-06-18
encode: extract can_emit_raw_unknown() for UNKNOWN_ENT/OBJ gating

informatimago:
The condition deciding whether an UNKNOWN_ENT/UNKNOWN_OBJ raw blob can be
re-emitted verbatim (same version && raw bits still held) was duplicated in
both switch cases. Extract it into a can_emit_raw_unknown() predicate so the
two paths can't drift apart. No behavior change (verified: example_2018
--as r2018 still round-trips with no object loss or wrong-size).

Addresses review feedback on #1286.

### DIFF
--- a/src/encode.c
+++ b/src/encode.c
@@ -6061,6 +6061,17 @@ dwg_encode_unknown_bits (Bit_Chain *restrict dat, Dwg_Object *restrict obj)
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


rurban (approved):
Yes, that's better

PR #1306, opened 2026-07-03
Make R2004+ decompression size cap configurable

### LINKED ISSUE #1305: Make the R2004+ decompression size cap configurable for very large DWG files
## Summary

LibreDWG currently has a hard-coded R2004+ decompression safety cap of `0x2f000000` (~790 MB) in `src/decode.c`. This is a useful default guard, but it also makes some valid, very large DWG files fail before callers have any way to opt into a higher limit.

I am not proposing to raise the default for everyone. The request is to keep the current default, but make the cap configurable so downstream tools can explicitly opt into a larger limit when they know they are processing unusually large files and have enough memory available.

## Real-world case

In a downstream CAD conversion pipeline, we encountered a large DWG whose `AcDb:AcDbObjects` compressed section metadata reports a decompression size above the current hard-coded cap. LibreDWG fails during package/open before entity-level recovery can run.

Example error:

```text
AcDb:AcDbObjects
num_sections: 35863
size: 0x3F79F8CF
ERROR: Invalid section AcDb:AcDbObjects count or max decompression size.
ERROR: Failed to read compressed AcDbObjects section
```

The DWG itself is much smaller on disk, but the decoded object section is large enough to exceed the built-in cap. This is expected for some drawings with many entities, blocks, hatches, or expanded CAD data.

## Desired behavior

- Keep the existing default cap (`0x2f000000`) unchanged.
- Allow callers/operators to opt into a larger cap explicitly.
- Invalid configured values should be ignored rather than making decoding less safe.

## Proposed configuration

The linked PR adds an environment variable:

```sh
LIBREDWG_MAX_DECOMP_SIZE=0x7fffffff
```

The value is parsed with base auto-detection, so both decimal and `0x...` hex values are accepted. If unset, LibreDWG keeps the current behavior.

## Related issue

This is related to large R2004+/R2018 compressed section handling, but not identical to #1294. That issue appears to involve section/page size accounting. This one is specifically about allowing the existing maximum decompression guard to be configured without changing the default.

BenLampson:
## Summary

This keeps the current R2004+ maximum decompression size guard unchanged by default, but allows operators to override it explicitly with an environment variable when processing unusually large DWG files.

The new variable is:

```sh
LIBREDWG_MAX_DECOMP_SIZE=0x7fffffff
```

The value is parsed with base auto-detection, so decimal values and `0x...` hexadecimal values are both accepted. Invalid values are ignored and the default remains in effect.

## Motivation

Downstream conversion services may encounter valid DWG files whose compressed `AcDb:AcDbObjects` section expands beyond the current hard-coded `0x2f000000` cap. In that case LibreDWG fails before callers can recover at the entity level.

This PR does not raise the default cap globally. It only makes the existing safety limit configurable for deployments that intentionally opt into a higher memory ceiling.

## Default behavior

- Default remains `0x2f000000`.
- If `LIBREDWG_MAX_DECOMP_SIZE` is unset, behavior should be unchanged.
- If the variable is invalid, LibreDWG logs a warning and keeps the default.

Closes #1305.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -72,6 +72,32 @@ static bool env_var_checked_p;
 // #undef LOG_POS
 // #define LOG_POS LOG_INSANE (" @%" PRIuSIZE ".%u\n", dat->byte, dat->bit);
 
+#define DEFAULT_MAX_R2004_DECOMP_SIZE 0x2f000000U
+
+static bool max_decomp_size_checked_p;
+static uint32_t max_r2004_decomp_size = DEFAULT_MAX_R2004_DECOMP_SIZE;
+
+static uint32_t
+get_max_r2004_decomp_size (void)
+{
+  if (!max_decomp_size_checked_p)
+    {
+      char *probe = getenv ("LIBREDWG_MAX_DECOMP_SIZE");
+      if (probe && probe[0])
+        {
+          char *endptr = NULL;
+          unsigned long value = strtoul (probe, &endptr, 0);
+          if (endptr != probe && endptr && *endptr == '\0' && value > 0
+              && value <= UINT32_MAX)
+            max_r2004_decomp_size = (uint32_t)value;
+          else
+            LOG_WARN ("Ignore invalid LIBREDWG_MAX_DECOMP_SIZE=%s", probe);
+        }
+      max_decomp_size_checked_p = true;
+    }
+  return max_r2004_decomp_size;
+}
+
 /*------------------------------------------------------------------------------
  * Private functions
  */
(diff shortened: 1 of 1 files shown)


rurban (commented):
Yes, but I'm tending to lift that hard limit, and rather check the dat->byte position vs filesize to bail out.

rurban:
Testing my idea in my PR #1307

BenLampson:
I think your solution is great rurban!
My env var change was just a compromise to keep original behavior, removing the hard limit directly is definitely the better route.
I’m closing this PR, thanks a lot for your input!

PR #1308, opened 2026-07-05
clamp negative num_hdl in 2ndheader before writing hdl vector

saddamr3e:
The SecondHeader keeps a per-slot handle backup in a fixed hdl[8] array, with num_hdl holding how many bytes are valid. num_hdl is a signed BITCODE_RCd, and while the decoder and encoder both loop with a num_hdl > 0 guard, the JSON writer goes through the default SUB_FIELD_VECTOR_N which casts the count to BITCODE_BL. A byte >= 0x80 read as int8_t becomes negative, and that cast turns it into a huge count, so json_section_2ndheader walks far past hdl[8]. The existing SUB_VALUEOUTOFBOUNDS next to it only rejects values above 8, never negatives, so nothing catches it. I hit this feeding a crafted R13-R2000 file through decode then --format JSON: UBSan reports index 8 out of bounds for BITCODE_RC[8]. Clamping a negative num_hdl to 0 right after it is read in the shared spec closes the read for every consumer while leaving valid 0..8 counts untouched.

### DIFF
--- a/src/2ndheader.spec
+++ b/src/2ndheader.spec
@@ -65,7 +65,19 @@ VERSIONS (R_13, R_2000) {
   VALUEOUTOFBOUNDS (num_handles, 14);
   REPEAT_F (num_handles, 14, handles, Dwg_SecondHeader_Handles)
   REPEAT_BLOCK
+      ENCODER {
+        // signed field, but used as unsigned vector count for hdl[8]:
+        // never write a negative count
+        if (_obj->handles[rcount1].num_hdl < 0)
+          _obj->handles[rcount1].num_hdl = 0;
+      }
       SUB_FIELD_RCd (handles[rcount1], num_hdl, 0); // max 8, the size
+      DECODER {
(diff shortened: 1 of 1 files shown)


rurban (changes requested):
This needs seperate read and write paths. Your patch is only for write, which should be wrapped in ENCODE {}

saddamr3e:
Split it the way you asked: decode clamps num_hdl after the read, encode clamps before the write in ENCODE {}.

Looking again though, 6d8292c already switched num_hdl to BITCODE_RC (unsigned), so 0xFF now reads as 255 and the >8 SUB_VALUEOUTOFBOUNDS catches it. That fixes the same json overread at the type level and makes a negative clamp here moot. Fine to close this as covered by that commit, or I can trim it down to just an encode-side >8 guard if you still want that on the writer.

rurban:
Yes, someone else detected it before, and I just fixed that.

PR #1312, opened 2026-07-17
in_dxf: fix crashes and aborts importing AutoCAD dynamic-block DXFs

tuxiasumari:
Importing a real AutoCAD 2018 DXF that uses dynamic blocks either segfaulted or aborted the whole conversion (`READ ERROR 0x800`). Chasing one 4.5 MB real-world pavement plan through gdb surfaced four independent bugs, fixed here:

## 1. `add_EVAL_Node`: 92-counter never reset per node
Each `ACAD_EVALUATION_GRAPH` node carries four 92 group codes. The `j` counter survived across nodes, so the second node's first 92 hit the `j > 3` branch and was misread as the start of the edge list, derailing into `Invalid DXF code 92 for EVALUATION_GRAPH` and aborting the import. Also, group 95 is the node's `nextid` (as dwg.spec documents); the importer was writing it into `edge_flags` and warning `!= 32` on valid files.

## 2. `add_EVAL_Edge`: heap overflow on first edge
`realloc (o->edges, o->num_edges * sizeof ...)` runs with `num_edges == 0` on the first iteration — a 0-byte allocation immediately written through `edges[0]`.

## 3. `add_AcDbBlockStretchAction`: unallocated indexes arrays (SEGV)
`hdls[i].indexes` and `codes[i].indexes` were filled without ever being allocated (`xcalloc` of the parent array leaves the pointers NULL) — reproducible SEGV at the `o->hdls[i].indexes[j]` write. The `codes` loop bound was also copy-pasted from `hdls[i].num_indexes`.

## 4. `add_FIELD`: AutoCAD childval layout rejected
AutoCAD writes each childval's data type as group **93** (not 90), with optional typed payload pairs (90 long, 140 double, 330 id, 1 string, 11/21/31 point), 300/302 format strings, a `304 ACVALUE_END` terminator, and the evaluated cache value under a 7 key. The strict 90-based reader aborted the whole file (`Invalid DXF code 93 for FIELD`). The childval block is now parsed tolerantly, accepting both this layout and the legacy 90-based one — one unexpected pair no longer kills the import.

## Testing
- The reproducer plan now converts with `dxf2dwg --as r2000` cleanly; the output DWG opens in ezdxf-based viewers (verified geometry against the source).
- All 254 unit tests pass on 0.14 + these fixes (`-Wno-error` for the pre-existing `add_test.c` const warnings under GCC 15).
- Sample file available privately (third-party project data).

Sibling PR from the same hunt: #1311. Found while building [IngeCAD](https://github.com/tuxiasumari/ingecad), which embeds LibreDWG as its DWG satellite.

### DIFF
--- a/src/in_dxf.c
+++ b/src/in_dxf.c
@@ -6212,8 +6212,10 @@ add_EVAL_Edge (Dwg_Object *restrict obj, Bit_Chain *restrict dat,
   while (pair != NULL && pair->code == 92)
     {
       i++;
+      // i+1, not num_edges: on the first edge num_edges is still 0 and
+      // realloc'ing to 0 bytes made the edges[0] write below a heap overflow
       o->edges = (Dwg_EVAL_Edge *)realloc (
-          o->edges, o->num_edges * sizeof (Dwg_EVAL_Edge));
+          o->edges, (i + 1) * sizeof (Dwg_EVAL_Edge));
       if (!o->edges)
         {
           o->num_edges = 0;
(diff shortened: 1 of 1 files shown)


rurban:
Too big. needs a CLA

tuxiasumari:

You reimplemented it in the 2026-07-26 series on master — for this one mainly `94c3fc28` (in_dxf: fix crashes importing AutoCAD dynamic-block DXFs). I verified the content is in `master` before closing, so there is nothing left to review here.

PR #1314, opened 2026-07-17
dxf: emit valid strings — caret-encode controls, embed before quote

tuxiasumari:
Sweeping a 1,385-file real-world DWG corpus (Peruvian civil-engineering drawings, r2000–r2018) through `dwg2dxf` produced 199 outputs that no DXF reader accepts. All traced to string emission in `out_dxf.c`; four related fixes:

## 1. `cquote`: caret-encode all C0 control characters
LF→`^J` and CR→`^M` are two cases of the general caret rule (`c + 0x40`). Decoder over-reads (XRECORD material XML, GEODATA coordinate-system tails) leave arbitrary control bytes (ETX, DC1, ...) inside strings, and a single raw byte < 0x20 desyncs the tag/value pair stream for every reader. Tab stays literal, as AutoCAD keeps it.

## 2. `dxf_fixup_string`: quote AFTER the utf8→TU→embed roundtrip
Quoting ran before `bit_utf8_to_TU` + `bit_embed_TU`; invalid UTF-8 (from the same over-reads) misdecodes into arbitrary code units, and embed re-emits code points < 0x20 as raw control bytes — undoing the quoting.

## 3. Never split a `\U+XXXX` escape across continuation chunks
The 250-char chunking could cut right after `\U+`, which crashes ezdxf (`int('', 16)`) and corrupts other readers. Chunks now break before a trailing backslash.

## 4. Continuation group codes for 3xx string fields
Continuations now use `dxf + 2` (301 → 303), matching AutoCAD's own GEODATA output, instead of a hardcoded 3.

Plus `dxf_fputs_escaped`, an escaping writer used for all chunked string output as the last line of defense.

## Results
- Corpus: 199 unreadable conversions → 10; overall success 83% → 96.6% (1150 → 1338 of 1385 files load in ezdxf with entities intact).
- All 254 unit tests pass on 0.14 + these fixes.
- Related ezdxf robustness note: a bare `\U+` without digits raises an uncaught `ValueError` there; will report separately to mozman.

Sibling PRs from the same hunt: #1311, #1312, #1313. Found while building [IngeCAD](https://github.com/tuxiasumari/ingecad).

### DIFF
--- a/src/out_dxf.c
+++ b/src/out_dxf.c
@@ -1322,15 +1322,14 @@ cquote (char *restrict dest, const size_t len, const char *restrict src)
   char *s = (char *)src;
   while ((s < send) && (c = *s++) && dest < dend)
     {
-      if (c == '\n' && dest + 1 < dend)
+      if (c > 0 && c < 0x20 && c != '\t' && dest + 1 < dend)
         {
+          /* caret-encode all C0 control chars, the general rule AutoCAD
+             uses: LF -> ^J, CR -> ^M, ETX -> ^C ... Raw control bytes
+             (seen in GEODATA coord_system_def tails) produce DXF every
+             reader rejects. Tab stays literal like AutoCAD keeps it. */
           *dest++ = '^';
-          *dest++ = 'J';
-        }
-      else if (c == '\r' && dest + 1 < dend)
-        {
-          *dest++ = '^';
-          *dest++ = 'M';
+          *dest++ = c + 0x40;
         }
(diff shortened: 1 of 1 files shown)


rurban:
Excellent but needs a CLA

tuxiasumari:

You reimplemented it in the 2026-07-26 series on master — for this one mainly `64c1e225` (dxf: emit valid control characters and fix string quoting order). I verified the content is in `master` before closing, so there is nothing left to review here.

PR #1315, opened 2026-07-17
dxf: sanitize partially decoded values — zero handles, MTEXT columns

tuxiasumari:
Two more findings from sweeping a 1,385-file real-world DWG corpus, both of the form "partial decode leaves garbage that makes the emitted DXF fatally invalid":

## 1. Objects written with handle 0
Partially decoded objects (LAYOUTs in the wild, plus the synthetic ENDBLK from `dxf_ENDBLK_empty`) get written as `5/0`. Readers reject handle 0 outright (ezdxf: `ValueError: Invalid handle 0`), losing the whole file over a filler object. `dxf_fixup_zero_handle` assigns a fresh unique handle at write time (monotonic on top of `dwg_next_handle`).

## 2. MTEXT embedded-object column garbage
Partial decodes leave `column_type` uninitialized (values like 256, 8908, 35141 seen across 6 files, with `column_width` at 1.3e+218). A 71 group > 2 crashes ezdxf's `ColumnType` enum and corrupts other readers' column parsing. Out-of-range values are now treated as "no columns" — the column block is skipped.

## Results
Fixed 7 of the corpus' remaining unreadable conversions (6× ColumnType, 1× handle 0). All 254 unit tests pass on 0.14 + the full patch series (this one applies independently of #1314).

Also observed, for the record (deeper decode-side issues, happy to share samples privately): files whose entities decode with massively duplicated low handles (#1..#2A), r2018 files where most objects come out as UNKNOWN_OBJ/UNKNOWN_ENT (Civil 3D proxies), and AC1009 (r11) files failing at `DWG_SENTINEL_R11_LAYER_BEGIN`.

Sibling PRs: #1311 #1312 #1313 #1314. From [IngeCAD](https://github.com/tuxiasumari/ingecad)'s corpus harness (`tools/dwg_bench.py`).

### DIFF
--- a/src/dwg.spec
+++ b/src/dwg.spec
@@ -2944,6 +2944,10 @@ DWG_ENTITY (MTEXT)
         FIELD_BD (extents_height, 43);
         // end redundant fields
 
+        // Partial decodes leave garbage here (256, 35141...); a 71 value
+        // > 2 crashes DXF readers (ezdxf ColumnType) — treat as "no columns"
+        if (FIELD_VALUE (column_type) > 2)
+          FIELD_VALUE (column_type) = 0;
         FIELD_BS (column_type, 71);
         if (FIELD_VALUE (column_type))
           {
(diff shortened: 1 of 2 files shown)


rurban (approved):
Looks good, I have to check. Thanks

PR #1317, opened 2026-07-19
in_dxf: five field misreads that abort or corrupt real-file imports

tuxiasumari:
Five independent `in_dxf` bugs found importing real AutoCAD/ezdxf-written DXFs (same corpus and methodology as #1311–#1315):

- **DIMENSION group 16/26/36 aborted the whole file.** The dimension-arc definition point is valid DXF on angular dimensions; subtypes whose DWG struct has no such field (`DIMENSION_ANG3PT` as written by AutoCAD/ezdxf) must skip it, not fatal-out.
- **BLOCKROTATEACTION's third connection point is 96/305**, not 94/303 (the writer emits dxf+vcount starting at 2, matching AutoCAD); expecting 94/303 rejected every real file carrying one.
- **A SEQEND's 330 may point at the owning BLOCK_RECORD** (ezdxf and AutoCAD's own DXF do this), not the INSERT/POLYLINE parent. Trusting it wrote `num_owned = <all entities of the block>` onto the BLOCK_HEADER with no matching entities array — an out-of-bounds crash later in `dwg_fixup_BLOCKS_entities`. Only real SEQEND parents qualify; anything else falls back to the backwards search.
- **2D point vectors lost every point's Y.** The generic point-VECTOR handler stored the first point's .x (code 10) but its .y (code 20, still j==0) fell through the `j > 0` branch, and 2D points never advanced `j`. A HATCH seed came in as (x, 0) — far outside the hatch — which AutoCAD/ODA reject ("Object improperly read <AcDbHatch>, Out of memory").
- **Setting only a point's .x leaked the previous field's .y/.z** (the `pt` local is reused across fields). E.g. an INSERT whose DXF has group 41 (xscale) but omits 42/43 got scale.y/z from the insertion point — giant streaked geometry. Read-modify-write instead.

Validated on 0.14 with the full unit-test suite (254 green) plus round-trip verification against ODA's ACAD2000 output as a strict oracle and visual checks in BricsCAD; the master port is line-identical and compile-checked (no autotools here for a full master `make check`).

### DIFF
--- a/src/in_dxf.c
+++ b/src/in_dxf.c
@@ -9359,9 +9359,27 @@ dxf_postprocess_SEQEND (Dwg_Object *restrict obj)
 
   loglevel = dwg->opts & DWG_OPTS_LOGLEVEL;
   LOG_TRACE ("dxf_postprocess_SEQEND:\n");
-  // r12 and earlier: search for owner backwards
-  if (dwg->header.from_version < R_13b1 && !owner
-      && !obj->tio.entity->ownerhandle)
+  // Writers like ezdxf (and AutoCAD's own DXF) set 330 of a SEQEND to the
+  // owning BLOCK_RECORD, not to the INSERT/POLYLINE parent. Trusting that
+  // owner wrote num_owned = <all entities of the block> onto the
+  // BLOCK_HEADER (a field it does have) with no matching entities array —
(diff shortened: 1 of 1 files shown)


rurban:
Excellent. But too big. Needs a CLA

rurban:
Fixed independently by myself in-between, thanks to your report. Dont want to wait for the CLA

PR #1318, opened 2026-07-19
HATCH: three spline-edge/path bugs that corrupt or drop hatch data

tuxiasumari:
Three HATCH bugs, all exposed by complex real-world hatches (pavement/plaza detail sheets) through the `dwg2dxf → dxf2dwg` r2000 round trip; each alone produces an unreadable hatch ("Object improperly read: <AcDbHatch>") or silently drops data:

- **dwg.spec (decoder, R2010+ spline edges): fit points/tangents must be read only when `num_fitpts > 0`.** AutoCAD writes the two tangents only for fit-point-defined spline edges; reading them unconditionally overran the object and the decoder silently dropped every remaining edge/path of the hatch, so `dwg2dxf` emitted counts that disagreed with the data (a 3-path/6-edge hatch came out as 2/3).
- **`in_dxf`: `has_derived` must OR every path flag's 0x4 bit** (that is how the decoder computes it); plain `=` kept only the last path's bit, so the encoder skipped `pixel_size` while the reader expected it → misaligned `num_seeds`.
- **`in_dxf`: a rational spline's knots were read as weights.** Knots (code 40, `num_knots` of them) precede the per-control-point weights (also code 40); keying only on `is_rational` sent every knot after the first to the weight branch, corrupting knots and overrunning the control-point array ("wrong l N control_points index", aborting the import with READ ERROR). Track the pending-knot count.

With these three, two previously-fatal real files (both fail on master) convert clean and open in BricsCAD/ODA. 254 unit tests green on 0.14; master port line-identical, compile-checked.

### DIFF
--- a/src/dwg.spec
+++ b/src/dwg.spec
@@ -4743,9 +4743,17 @@ DWG_ENTITY (HATCH)
                         {
 #define seg segx[rcount2]
                           SUB_FIELD_BL (seg, num_fitpts, 97);
-                          FIELD_2RD_VECTOR (seg.fitpts, seg.num_fitpts, 11);
-                          SUB_FIELD_2RD (seg, start_tangent, 12);
-                          SUB_FIELD_2RD (seg, end_tangent, 13);
+                          // AutoCAD writes the fit points and the two
+                          // tangents only when the spline edge was defined by
+                          // fit points (num_fitpts > 0). Reading the tangents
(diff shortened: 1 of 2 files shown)


rurban (approved):

rurban:
The DWG's would be very interesting also

PR #1319, opened 2026-07-19
bits: r2000 TV length excludes the trailing NUL; fix bit_write_DD mode selection

tuxiasumari:
Two low-level encoder bugs that made LibreDWG-written r2000 DWGs unreadable or subtly corrupt for AutoCAD/BricsCAD/ODA:

- **`bit_write_TV`: the trailing NUL in the TV length is an r2004+ convention.** Pre-r2004 lengths equal `strlen` — ODA's ACAD2000 output confirms ("SOLID" is written as 5, not 6). Writing the NUL into r2000 made *every string one byte long*, shifting the object body; the first victim in a real file is typically a HATCH ("Object improperly read: <AcDbHatch>" in BricsCAD/ODA). Gate on `>= R_2004`. The reader side already tolerates both. Companion test asserts both layouts (r2000 @4.2, r2004 @5.2).
- **`bit_write_DD` aliased the pointer, not the value, and compared the wrong words.** `(const uint16_t *)&uchar_value` takes the address of the local *pointer variable*, so the DD mode (how many bytes to emit) was chosen from stack garbage; and the comparison used the LOW words while `bit_read_DD` keeps the HIGH bytes of the default (word 3 for code 2, words 2+3 for code 1). Partial-mode writes therefore corrupted values (e.g. an INSERT yscale of 1 came back as the insertion-point Y). Alias the value bytes and compare the high words so encode matches decode.

Found via bit-position diffing of `dwgread -v5` output against ODA's ACAD2000 files as a strict oracle (same corpus as #1311–#1315). With these (plus the import fixes in the sibling PRs), previously-rejected LibreDWG r2000 output opens and renders correctly in BricsCAD. 254 unit tests green on 0.14, including the new TV cases; master port line-identical, compile-checked.

### DIFF
--- a/src/bits.c
+++ b/src/bits.c
@@ -1300,15 +1300,23 @@ bit_write_DD (Bit_Chain *dat, double value, double default_value)
       const unsigned char *uchar_value = (const unsigned char *)&value;
       const unsigned char *uchar_default
           = (const unsigned char *)&default_value;
-      const uint16_t *uint_value = (const uint16_t *)&uchar_value;
-      const uint16_t *uint_default = (const uint16_t *)&uchar_default;
-      // dbl: 7654 3210, little-endian only
-      // check the first 2 bits for eq
-      if (le16toh (uint_value[0]) == le16toh (uint_default[0]))
+      // Alias the DOUBLE bytes, not the address of the pointer variable:
+      // (const uint16_t *)&uchar_value compared two stack addresses, so the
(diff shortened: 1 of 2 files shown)


rurban:
Good, but too big

tuxiasumari:

You reimplemented it in the 2026-07-26 series on master — for this one mainly `ac644ec1` (bits: fix r2000 TV length and bit_write_DD mode selection). I verified the content is in `master` before closing, so there is nothing left to review here.

PR #1320, opened 2026-07-19
in_dxf/dwg: handle & ownership fixes so conformant readers accept our r2000 output

tuxiasumari:
Five ownership/handle-semantics bugs. Each one makes AutoCAD/BricsCAD/ODA warn on or reject a DWG written from imported DXF; together they were the difference between "opens with recovery errors" and "opens clean" for a 92k-entity cadastre:

- **SORTENTSTABLE sort handles must not register in `object_map`.** sort_ents are code-0 *ordering* handles that intentionally alias real entity handles; passing `obj` to `dwg_add_handleref` registered each one in the object map, colliding with the true entity (a real cadastre emitted 3,040 "Duplicate handle" errors) and corrupting handle resolution → invalid block owners downstream. Pass NULL.
- **`SORTENTSTABLE.block_owner` was the extension DICTIONARY, not the BLOCK_HEADER.** An empty block-scoped table has no explicit 330 for it in DXF and the importer misread the object owner. Derive it from the ownership chain (owner of the owning dictionary); non-empty tables with a distinct explicit block_owner are trusted as-is.
- **Regular block members were imported with entmode 3** (model space, no owner in stream). AutoCAD/ODA write entmode 0 (explicit owner = block record) for block members — with 3, conformant readers resolve the BlockBegin owner to *Model_Space and warn "BlockBegin owner Id is invalid" on every block.
- **A bound/local block (e.g. `A$C…`) kept a stray xref bit** in BLOCK_RECORD flag 70; with no xref path it must be local, or readers render whole layers as unresolved "Xref" placeholders.
- **`dwg_set_next_objhandle` honored a stale `next_hdl` without checking `object_map`.** When a table record is added at encode time after DXF import (the "LibreDWG"/ACAD_MLEADERVER APPIDs), it could land on a handle already owned by e.g. an LTYPE → ambiguous handle map → "AcDbRegAppTableRecord can't be cast to AcDbEntity". Reuse the requested handle only if free, else fall through to a fresh handseed.

Verified field-by-field against ODA's ACAD2000 output (entmode distribution, owner warnings 10→0 / 45→0 on the bench files) and visually in BricsCAD. 254 unit tests green on 0.14; master port line-identical, compile-checked.

### DIFF
--- a/src/dwg.c
+++ b/src/dwg.c
@@ -3503,11 +3503,30 @@ dwg_set_next_objhandle (Dwg_Object *obj)
   BITCODE_RLL seed;
   if (dwg->next_hdl)
     {
-      obj->handle.value = dwg->next_hdl;
-      dwg_set_handle_size (&obj->handle);
-      hash_set (dwg->object_map, obj->handle.value, (uint64_t)obj->index);
+      // Only honor the requested handle if it is still free. A stale next_hdl
+      // (e.g. left over from DXF import when a table record is added later at
+      // encode time, like the "LibreDWG"/ACAD_MLEADERVER APPIDs) would
+      // otherwise collide with an existing object and produce a DWG whose
+      // handle map is ambiguous ("... TableRecord can't be cast to
+      // AcDbEntity" in AutoCAD/BricsCAD/ODA). On collision, fall through to a
+      // fresh handseed.
+      uint64_t existing = dwg->object_map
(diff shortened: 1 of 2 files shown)


rurban:
Excellent! But too big without CLA

rurban:
Fixed independently by myself, thanks to your description.

PR #1321, opened 2026-07-19
in_dxf/dynapi: MTEXT and DIMENSION import fidelity (giant text, distorted dims)

tuxiasumari:
Four fixes for the two most visible symptoms on DXF-imported drawings re-saved as r2000 — MTEXT rendering ~14× too big and distorted dimensions:

- **dynapi: MTEXT `rect_width` is 41 and `rect_height` is 46** (they were 40/41, clobbering `text_height`'s 40). Root cause in `gen-dynapi.pl`: a field re-read in a later embedded/annotative context with a different group code overwrote the canonical mapping — keep the FIRST dxf code per field (generator fixed + `dynapi.c` regenerated for MTEXT).
- **in_dxf: the MTEXT R2018 "Embedded Object" (code 101) clobbered the entity.** Its fields reuse main-entity codes (40=rect_width, 41=rect_height, 10/11 points…); the importer set `in_embedobj` but never checked it, so e.g. embedded 40=34.9 landed on `MTEXT.text_height` (2.5) → giant labels on almost every converted sheet. Skip the embedded object's fields (r2000 output drops it anyway).
- **in_dxf: MTEXT flow direction / line-spacing defaults.** These live only in the DWG; ezdxf omits them and the calloc zeros (flow 0, lspace style/factor 0) mis-lay-out the text in AutoCAD/BricsCAD. Apply the defaults ODA writes; any codes the DXF does carry overwrite them.
- **in_dxf: DIMENSION common defaults.** Same pattern: `ins_scale (0,0,0)`, lspace 0/0, `act_measurement 0` shipped from calloc; AutoCAD/ODA mis-scale or reject those. Set ins_scale (1,1,1), lspace 1/1.0, act_measurement -1 ("recompute on open"), matching ODA.

Verified with `dwgread -v5` bit-diffs against ODA's ACAD2000 output and side-by-side rendering in BricsCAD (the giant-text and dim symptoms disappear; normal MTEXT unchanged). 254 unit tests green on 0.14; master port line-identical, compile-checked.


tuxiasumari:

You reimplemented it in the 2026-07-26 series on master — for this one mainly `af364d4c` (gen-dynapi: keep first DXF group code per field to fix MTEXT rect size). I verified the content is in `master` before closing, so there is nothing left to review here.

PR #1322, opened 2026-07-19
Convert UTF-8 DXF text to the target codepage for pre-r2007 (mojibake)

tuxiasumari:
Importing a UTF-8 DXF into r2000 stored **raw UTF-8 bytes in single-byte codepage TV fields**; readers decode them as the drawing codepage, so `Á` (UTF-8 C3 81) became `Ã` + a control char, and `Ñ/Ó/É` split into two wrong glyphs on every text of every converted Spanish-language drawing.

Three code paths skipped the UTF-8→codepage conversion for `< R_2007`, each fixed to convert via `bit_utf8_to_TV` (characters not representable in the codepage become `\U+XXXX` escapes, matching AutoCAD):

- **`dwg_add_u8_input`**: the codepage branch existed but was `#if 0`-ed — enabled and completed (growing buffer, UTF-8 fallback, keeping the R12 32-byte name guard).
- **`dynapi_set_helper`**: the "ascii" pre-r2007 branch just copied bytes. It now converts when `is_utf8 && codepage > 1` (a `codepage` parameter threads through its five callers); `is_utf8 == 0` (DWG decode) still copies as-is. This covers TEXT and every TV field set through dynapi.
- **`MTEXT.text`** has its own code-1/3 chunk assembler that `strdup`-ed raw UTF-8 — added a `dxf_postprocess_MTEXT` doing the same conversion. (Field type "T" and the MTEXT special handler take different paths; fixing one does not fix the other.)

Verified byte-level via ODA: É=C9, Ñ=D1, Á=C1, Ó=D3 in CP1252 output, round-trip identical; r2007+ (TU) untouched; hatches and the rest of the bench files unaffected. 254 unit tests green on 0.14; master port line-identical, compile-checked.

### DIFF
--- a/src/dwg_api.c
+++ b/src/dwg_api.c
@@ -24771,32 +24771,11 @@ dwg_add_u8_input (Dwg_Data *restrict dwg, const char *restrict u8str)
     }
   else
     {
-      // TODO Encode unicode to \U+... bit_utf8_to_TV. codepage conversions
-#if 0
-      int size = 1024;
-      char *dest = (char *)malloc (size);
-      char *tgt = bit_utf8_to_TV (dest, u8str, size, strlen(u8str), 0,
-                                  dwg->header.codepage);
-      if (!dest)
-        {
-          LOG_ERROR ("%s: Out of memory", __FUNCTION__);
-          return NULL;
-        }
-      while (!tgt)
-        {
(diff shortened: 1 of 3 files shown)


rurban:
Fixed by myself in between

PR #1326, opened 2026-07-26
Fix encoding in AC1006

michal-josef-spacek:
There were bad size in APPID.


michal-josef-spacek:
Test is possible via run of `examples/dwgadd_test.sh` on both versions. Original has ERROR.

PR #1327, opened 2026-07-26
encode: free imported class dxfname before replacing it

nameloCmaS:
This is part of the DXF importer allocation-ownership cleanup found while reviewing Fuzzing Action LeakSanitizer reports.

Imported DXF objects own their duplicated dxfname. Free that string before replacing it with a class alias or class dxfname during encode-time class lookup.

This does not overlap the parsing or semantics scope of PR #1312-#1322.

* src/encode.c (dwg_encode_get_class): Free IN-mode object dxfname before assigning a new duplicate.


rurban:
fixing indxf leaks is a huge task, I left aside for some years. Thanks

PR #1358, opened 2026-08-06
decode: r2004 sections whose declared size exceeds the page estimate (fixes #1294)

### LINKED ISSUE #1294: AC1032 (r2018): large compressed AcDbObjects section fails to decompress - "Invalid section size 9918156 > 333 * 29696", num_objects: 0
## Summary

A specific AC1032 (r2018) DWG decodes to `num_objects: 0`: `dwgread` fails to read the section info (first `SummaryInfo`, then the compressed `AcDbObjects` section). The file is valid, ODA/RealDWG-based tools read it fully (~76,360 entities, ~40 layers).

This is **not** a general r2018 problem and **not** size-related: 12 other r2018 files from the same client and CAD producers, including larger ones (up to 4.4 MB / ~90k objects), decode correctly with the very same `dwgread` build. Only this one file fails, so it looks like a file-specific structural edge case in the R2004+ section-info reader.

## Versions tested (both fail identically)

- `libredwg-0.13.4-win64` (stable release)
- `libredwg-0.13.4.8317-win64` (nightly, 2026-06-19), i.e. **after** the compressor/encoder rewrite mentioned in #918 and the fixes in #1124 / #948

## Same dwgread build, other r2018 files from the same source (all OK)

| file size | objects decoded | result |
|-----------|-----------------|--------|
| 0.3–1.8 MB (×9) | 3k–43k | OK |
| 3.6 MB | 90,896 | OK |
| 4.4 MB | 71,808 | OK |
| **4.0 MB (this file)** | **0** | **FAIL** |

## File / reproducer

- Version code: **AC1032 (r2018)**
- Size: ~4.2 MB, ~76k entities across ~40 layers (per ODA-based readers)

The triggering file is **confidential data and cannot be shared** (publicly or privately). To work around this, I'm glad to help reproduce without exposing the drawing:

- I can post the **full `dwgread -v4` structural log** (section/page tables, not drawing content).
- I can **run any debug build or patch you provide** against the file on my side and report the output.
- I can try to **produce a non-confidential synthetic r2018 file** large enough to hit the same multi-page `AcDbObjects` path, if you can hint at the size/structure threshold that triggers it.

## Actual output (`dwgread -v2 -O JSON`)
```
This file's version code is: AC1032 (r2018)
ERROR: Skip section AcDb:SummaryInfo with max decompression size 0x200 > 0x100
ERROR: Invalid section AcDb:SummaryInfo count or max decompression size. Sections: 1, Max size: 0
ERROR: Failed to read uncompressed SummaryInfo section
ERROR: Invalid section AcDb:AcDbObjects size 9918156 > 333 * 29696
ERROR: Failed to read compressed AcDbObjects section
ERROR: Invalid num_segidx
num_objects: 0
Warning: Object handle not found 16/0x10 in 0 objects of max 0x1C22A01D handles
... (33 such "in 0 objects" warnings)
```

## Analysis

`9,918,156 > 333 * 29,696` fails by 29,388 bytes (≈ one 29,696-byte page). The section seems to require **334** pages but only **333** are accounted for, suggesting an off-by-one / page-count miscomputation in the R2004+ section-info reader for large r2018 sections. This looks related to the previously fixed #1124 / #948 / #918, but those fixes did not cover this large-section case.

tuxiasumari:
Fixes #1294.

Two AC1032 files here decoded to `num_objects: 0` with

```
ERROR: Invalid section AcDb:AcDbObjects size 4678786 > 156 * 29696
ERROR: Failed to read compressed AcDbObjects section
```

although every page in the section is fine — ODA File Converter and BricsCAD
read the same drawings completely. A 1 MB public reproducer is attached to
#1294.

## First layer: the buffer is sized from an estimate

`num_sections * max_decomp_size` is an upper-bound *estimate*. Each page carries
its own `StartOffset` into the decompressed buffer, so what the buffer actually
has to span is `info->size`, the declared decompressed total. On these files the
estimate lands **1 to 3 pages short** of `info->size`, and the whole drawing is
refused rather than the buffer being grown:

| file | `info->size` | `num_sections` × 29696 | short by | pages |
|---|---:|---:|---:|---:|
| #1294's (not attached) | 9 918 156 | 9 888 768 | 29 388 | 0.99 |
| `cerco-perimetrico` | 4 678 786 | 4 632 576 | 46 210 | 1.56 |
| `planos-constructivos-A` | 13 535 788 | 13 511 680 | 24 108 | 0.81 |
| `planos-constructivos-B` | 22 261 264 | 22 182 912 | 78 352 | 2.64 |
| a file that passes | 80 910 | 89 088 | — | −0.28 |

Now whichever of the two is larger gets allocated. The 790 MB cap and the
64-bit overflow guard right above it are untouched, and `info->size < 0` is
still rejected.

## Second layer: the per-page guard uses the wrong size

This is what actually loses the section. The guard deciding whether a page fits
compares against the section-wide maximum instead of that page's own
decompressed size:

```c
if (info->compressed == 2 && bytes_left > 0
    && es.fields.address <= max_decomp_size
    && es.fields.address + info->max_decomp_size <= max_decomp_size)
```

The last page of a section is normally shorter, so `address + max_decomp_size`
overshoots the buffer while `address + page_size` fits comfortably. And when
that guard fails for a **compressed** section, control falls into the `else`
branch, whose very first condition is `info->compressed == 2` — so it returns
`"Some section size or address out of bounds"` and **156 good pages are thrown
away because of the last one**.

The `else` branch already computes the real size, eight lines below:

```c
const BITCODE_RL size = MIN ((BITCODE_RL)(info->size - es.fields.address),
                             es.fields.page_size);
```

The guard now uses the same expression, so both paths agree.

This also explains why only *some* r2018 files fail. When the page estimate is
generous there is slack and the guard passes by luck; when the pages are nearly
full there is none, and the final page kills the section every time. Which fits
the table above: the failures overshoot by under three pages, the file that
passes sits comfortably below.

It also argues against the deleted-objects theory from the issue thread:
`cerco-perimetrico` has `HANDSEED` `0x1313` = 4883 against ~2222 live entities,
a 2.2× ratio — an ordinary handle history — and it still failed.

## Verification

Against ODA File Converter 25.6, modelspace entities, same drawings:

| drawing | before | after | ODA |
|---|---:|---:|---:|
| `cerco-perimetrico.dwg` | **0** | **2222** | 2222 |
| `planos-constructivos-A.dwg` | **0** | **26583** | 26583 |
| `planos-constructivos-B.dwg` | **0** | **26583** | 26583 |
| control (passes the size check) | 1108 | 1108 | — |

Exact matches, entity for entity.

- `make check`: **270 PASS / 0 FAIL / 0 SKIP**, identical to the unpatched build.
- All **146** DWGs in `test/test-data` and `programs/` re-converted with and
  without the patch: **143 byte-identical, 0 worse**. (The 3 that differ are the
  `JUMP` fix in #1353, which the same build tree carries.)
- A 190-drawing corpus sweep, stock vs patched: **OK 160 → 166, 0 regressions**.

Built and tested on Ubuntu 26.04, gcc 15.2.0, on top of `e405fcff`
(= tag `0.14.8556`), ezdxf 1.4.4 and ODA File Converter 25.6 for the reference
side.

Happy to run any variation against the three files if you would rather bound the
page differently — @SimonSAMPERE can test his own file too, which the attached
reproducer does not replace.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -2061,6 +2061,22 @@ read_2004_compressed_section (Bit_Chain *dat, Dwg_Data *restrict dwg,
   {
     uint64_t max_decomp_size64
         = (uint64_t)info->num_sections * (uint64_t)info->max_decomp_size;
+    /* num_sections * max_decomp_size is only an upper-bound *estimate* of the
+       decompressed section. Each page carries its own StartOffset into the
+       buffer, so what the buffer really has to span is info->size, the declared
+       decompressed total. On some r2018 files the estimate comes out 1-3 pages
+       short of info->size; rejecting those loses the whole drawing
+       (num_objects: 0) even though every page is fine. Allocate whichever is
+       larger and keep the sanity cap. */
+    if (info->size > 0 && (uint64_t)info->size > max_decomp_size64)
+      {
+        LOG_TRACE ("Section %s: declared size %" PRId64 " exceeds the %"
+                   PRIu64 " estimate (" FORMAT_RL " * " FORMAT_RL
+                   "), allocating the declared size\n",
+                   info->name, info->size, max_decomp_size64,
+                   info->num_sections, info->max_decomp_size);
+        max_decomp_size64 = (uint64_t)info->size;
+      }
     if (max_decomp_size64 == 0 || max_decomp_size64 > 0x2f000000) // 790Mb
       {
         LOG_ERROR ("Invalid section %s count or max decompression size. "
@@ -2070,12 +2086,9 @@ read_2004_compressed_section (Bit_Chain *dat, Dwg_Data *restrict dwg,
       }
     max_decomp_size = (uint32_t)max_decomp_size64;
   }
-  if (info->size > (int64_t)info->num_sections * (int64_t)info->max_decomp_size
-      || info->size < 0)
+  if (info->size < 0)
     {
-      LOG_ERROR ("Invalid section %s size %" PRId64 " > %u * " FORMAT_RL,
-                 info->name, info->size, info->num_sections,
-                 info->max_decomp_size);
+      LOG_ERROR ("Invalid section %s size %" PRId64, info->name, info->size);
       return DWG_ERR_VALUEOUTOFBOUNDS;
     }
   LOG_HANDLE ("Alloc section %s size %" PRIu32 "\n", info->name,
@@ -2164,9 +2177,19 @@ read_2004_compressed_section (Bit_Chain *dat, Dwg_Data *restrict dwg,
       LOG_INSANE ("bytes_left:               %ld\n", bytes_left);
 
       // check if compressed at all
+      /* Bound the page by what it actually decompresses to, not by the
+         section-wide maximum: the last page is normally shorter, so
+         address + max_decomp_size overshoots the buffer while
+         address + page_size fits. When that happens for a *compressed*
+         section, control falls into the else branch below, which rejects it
+         outright on `info->compressed == 2` and loses the whole section. The
+         else branch already computes the real size this way. */
       if (info->compressed == 2 && bytes_left > 0
           && es.fields.address <= max_decomp_size
-          && es.fields.address + info->max_decomp_size <= max_decomp_size)
+          && es.fields.address
+                     + MIN ((BITCODE_RL)(info->size - es.fields.address),
+                            es.fields.page_size)
+                 <= max_decomp_size)
         {
           size_t orig_size = dat->size;
           dec.byte = es.fields.address;


SimonSAMPERE:
@tuxiasumari I tested your patch against my file : it decodes. Thanks for digging this out.

I built `56a05f98` (the parent of your `e4b7515`) and applied the PR diff on top, so `src/decode.c` is the only variable between the two runs and it also confirms the patch still applies cleanly.

```
before:  ERROR: Invalid section AcDb:AcDbObjects size 9918156 > 333 * 29696
         num_objects: 0                         minJSON:     51,776 bytes
after:   num_objects: 83917                     minJSON: 71,804,891 bytes
```

Your prediction was exact: 9,918,156 vs 333 × 29,696, short by 29,388 bytes = 0.99 pages.

I then ran both builds over my whole sample : 92 drawings (50 r2013, 21 r2010, 13 r2018, 6 r2007, 1 r2004, 1 r2000), so 91 of them go through `read_2004_compressed_section`. One file fixed, no new failures, and the minJSON output of the **other 91 is byte-identical** between the two builds. I did not run `make check`, so this is only a non-regression check on my data.

tuxiasumari:
Two notes, since this one has been sitting a while.

**It has been verified by the reporter.** @SimonSAMPERE closed the loop on #1294 on 2026-08-24: *"Solved by #1358 according to my tests"* — on his own file, not mine. That issue had been open since June for want of a shareable reproducer.

**The one red check is not a failure.** `ppc64-big-endian` reads:

```
2026-08-11T15:43:20Z ##[error]The operation was canceled.
```

The job was cancelled at 25m, not failed — there is no assertion, no test result, no compiler diagnostic in its log. That run is from 2026-08-11, and the same job passes on master (run 33048734338). The other 19 checks pass, big-endian `mips32` among them.

Happy to rebase or to push an empty commit if you would rather see the whole matrix green before merging.

Issue #1356, opened 2026-08-06
dwg2dxf: duplicate handles in the emitted DXF — a GROUP takes handle 2 and LAYOUT 'Model' becomes unloadable


tuxiasumari:
`dwg2dxf` writes DXF files in which several objects share a handle. When one of
those collisions lands on handle `2` — the `*Model_Space` `BLOCK_RECORD` — the
whole file becomes unloadable, because `LAYOUT 'Model'` then resolves to
whatever object was written last with that handle. ezdxf 1.4.4, both strict and
`recover`:

```
DXFStructureError: expected BLOCK_RECORD(#2) for layout 'Model' has invalid entity type: GROUP
```

Two sample files (AC1021 / R2007), shared by their author for this report:
**[handle-collision-dwgs.zip](https://github.com/tuxiasumari/libredwg/releases/download/bugfiles-2026-08-06/handle-collision-dwgs.zip)**

| file | bytes | sha256 |
|---|---:|---|
| `sedapar.dwg` | 11282880 | `e81562acd08d02bf5e517acd7434c98511f93d8106fedf132941dc15fb4bdd8d` |
| `cofopri.dwg` | 3976928 | `c93abe4a0c8a73df706be2c4e48882251710cef6640f375352f621c3d3990665` |

## Reproduce

```sh
programs/dwg2dxf -o sedapar.dxf sedapar.dwg
echo $?          # 0
python3 -c "import ezdxf; ezdxf.readfile('sedapar.dxf')"
                 # DXFStructureError: expected BLOCK_RECORD(#2) for layout 'Model' ...
```

## The collisions

Duplicate handles in the emitted DXF (same handle, group code 5, on more than
one object):

**`sedapar.dxf` — 6 collisions**

```
handle 2   ->  BLOCK_RECORD (*Model_Space)  +  GROUP      <-- this one breaks the file
handle 1   ->  TABLE                        +  ACDBPLACEHOLDER
handle 3   ->  DICTIONARY                   +  TABLESTYLE
handle 5   ->  BLOCK                        +  ACDBPLACEHOLDER
handle 6   ->  LAYER                        +  LONG_TRANSACTION
handle B9  ->  DICTIONARY                   +  ACDBPLACEHOLDER
```

**`cofopri.dxf` — 34 collisions**, mostly `TABLE`/`LTYPE`/`LAYER` against
`LAYOUT`:

```
handle 1B  ->  TABLE  + LAYOUT
[...]

PR #1363, opened 2026-08-06
decode_r2007: Reed-Solomon decode uncompressed data pages too

### LINKED ISSUE #1361: dwg2dxf: LWPOLYLINE point arrays desynchronise mid-list — 14% of vertices come out near DBL_MAX
`dwg2dxf` emits coordinates around 1e300 for a large share of `LWPOLYLINE`
vertices on some r2007 drawings. The point array reads correctly up to an
arbitrary index and then desynchronises, so the rest of that polyline comes out
as arbitrary bit patterns reinterpreted as doubles.

Reproducers are already attached to #1356:
**[handle-collision-dwgs.zip](https://github.com/tuxiasumari/libredwg/releases/download/bugfiles-2026-08-06/handle-collision-dwgs.zip)**
— `sedapar.dwg` (AC1021, 11 MB) is the worst of them, `cofopri.dwg` shows it too.

## Scale of it

Coordinates whose absolute value exceeds 1e12, outside the `HEADER` section,
same DWG through both converters:

| drawing | `dwg2dxf` | ODA File Converter 25.6 | worst value |
|---|---:|---:|---|
| `sedapar.dwg` | **45 316** | 40 | `1.791444824488772E+308` |
| `yanaquihua.dwg` | 452 | 8 | `1.78823489631404E+308` |
| `cofopri.dwg` | 185 | 6 | `6.720368213839343E+301` |
| a clean r2018 file, for contrast | 6 | 6 | `1E+20` (the `$EXTMIN` sentinel) |

Counting only modelspace `LWPOLYLINE` vertices in `sedapar.dwg`:

```
dwg2dxf   231138 points,  33188 corrupt   (14.4%)
ODA      1026048 points,      0 corrupt
```

Note the worst values sit just under `DBL_MAX` (1.7976931348623157e308), which is
what arbitrary bytes reinterpreted as a double tend to look like.

## The shape of the failure

Each affected polyline is fine for a while and then goes bad from one index
onwards — the index differs per entity, so it is not a fixed field or a constant
offset:

```
handle 11D   2541 points   valid to 566,   corrupt from 567
handle 124   8049 points   valid to 2521,  corrupt from 2522
handle 1F5   2541 points   valid to 1478,  corrupt from 1479
```

And the first bad point often keeps a plausible X while Y is already gone, which
looks like the bit position slipping mid-vertex rather than a whole point being
read from the wrong place:

```
valid    (247010.8, 8189001.4)          <- proper UTM
corrupt  (246616.7, -1.076384413476985e+228)
next     (246616.6, -1.651044677555219e+195)
```

`HATCH` shows it too in the same file (186 coordinates), and `LAYOUT` objects in
`cofopri.dwg` (87), so it is probably not specific to the `LWPOLYLINE` reader —
more likely a bit-stream desynchronisation that whatever entity happens to follow
inherits.

## Not the same as the object-map bug in #1355

Worth separating, since both produce nonsense numbers: #1355/#1360 was the object
map mis-parsing a modular char, and the drawing attached there
(`frontal.dwg`) comes out with **6** absurd coordinates after that fix — all of
them the normal `$EXTMIN`/`$EXTMAX` sentinels. So that path is clean now, and
this is something else.

The drawings above also hit #1356 (duplicate handles) and #1357/#1359 (the writer
abandoning a layout at the first unresolvable reference), and `sedapar.dwg` logs
~2000 `Invalid class index` errors, so its decode is unhealthy in more than one
way. Whether those share a root cause with this I could not tell from the
outside.

## Why it matters downstream

A single vertex at 1e301 is enough to make a viewer frame the whole drawing
inside an astronomically large box, at which point every real entity is smaller
than a pixel — a blank canvas holding a complete drawing. We ended up filtering
by the drawing's own `$EXTMIN`/`$EXTMAX` to survive it, which works but is
papering over the data.

Observed at `e405fcff` (= tag `0.14.8556`), Ubuntu 26.04, gcc 15.2.0,
ezdxf 1.4.4, ODA File Converter 25.6 for the reference side. Happy to run debug
builds against these files.

tuxiasumari:
`read_data_section()` reads "not compressed" as "not encoded", and copies such a
page straight out of the file buffer:

```c
      dat->byte = page->offset;
      // only if compressed. TODO: Isn't there a compressed flag as with 2004+?
      // theoretically the sizes could still be the same.
      if (section_page->comp_size != section_page->uncomp_size)
        {
          error = read_data_page (dat, &decomp[section_page->offset], ...);
        }
      else
        {
          ...
          memcpy (&decomp[section_page->offset], &dat->chain[dat->byte],
                  section_page->uncomp_size);
        }
```

Reed-Solomon coding, though, is independent of compression — and
`read_data_page()` already treats the two separately: it undoes the RS coding
first, and only then decompresses, `if (size_comp < size_uncomp)`. So an
uncompressed page can still be RS coded, and that `memcpy` hands the caller the
RS codewords: payload interleaved with parity.

## How the file shows it

An 11 MB r2007 drawing here has `AcDb:AcDbObjects` in 181 pages, **136 of them
stored uncompressed**. Splitting its object map by page kind:

| pages | objects | fail | |
|---|---:|---:|---:|
| compressed (45) | 10586 | 0 | 0.0% |
| uncompressed (136) | 2259 | 2183 | **96.6%** |

Not one object in a compressed page failed; almost none in an uncompressed page
survived. The failures are the usual symptoms of reading from the wrong bytes —
`Invalid object type 37521, only 65 classes`, `Invalid class index -500 < 0`,
`Invalid bitsize 1866445698 > 243880`. Shannon entropy of the affected byte
ranges is **7.87** bits/byte against **6.39** in the intact ones, which is what
parity interleaved with data looks like.

The 2183 lost objects are **2232 LWPOLYLINEs** plus a handful of
CIRCLE/MTEXT/TEXT/LINE, and their handles are still listed in `*Model_Space`'s
entity array, so they surface downstream as 2258 references that resolve to
nothing.

## Telling the two page kinds apart

The section page record has no compression flag — that is exactly what the TODO
in the comment asks about — so `page->size` is the only evidence. An RS coded
page is `ceil(comp_size / 0xFB) * 0xFF` rounded up to 32; a verbatim page is
only rounded up to 32, and comes out smaller. Over all 192 pages of that file:

| | pages | `page->size == RS size` |
|---|---:|---|
| compressed | 51 | yes, all |
| uncompressed | 137 | yes, all |
| uncompressed | 4 | no — `SummaryInfo`, `Preview`, `AppInfo`, `FileDepList` |

Those last four really are verbatim: read that way they give
`LASTSAVEDBY: "Nestor"`, `appinfo_name: "AppInfoDataList"`, `version:
"17.0.54.0"`, sane `TDCREATE`/`TDUPDATE`. So the patch takes the RS path only
when `page->size` matches the coded size exactly, and anything else keeps
today's verbatim copy. Compressed pages are untouched either way.

If you know the real discriminator I would rather use it than a size match — but
the size match is a necessary condition, not a guess: a page too small to hold
the codewords cannot be RS coded.

## Result

Against ODA File Converter 25.6 on that drawing:

| | before | after | ODA |
|---|---:|---:|---:|
| modelspace entities | 8588 | **10847** | 10847 |
| `LWPOLYLINE` vertices | 231138 | **1026048** | 1026048 |
| vertices near `DBL_MAX` | 33188 | **0** | 0 |
| coordinates above 1e12 | 45316 | **0** | 40 |

**This also closes #1361**, which I filed for `LWPOLYLINE` point arrays
"desynchronising mid-list". They were not desynchronising — they were being read
out of RS codewords. Sorry for the misdirection there; I will note it in the
issue.

## Verification

- `make check`: **270 PASS / 0 FAIL / 0 SKIP**, unchanged. That includes the two
  unit tests over this area, `decompress_r2007` and `read_data_section: rejects
  out-of-bounds page->offset`; both still pass.
- All **146** DWGs in `test/test-data` and `programs/` converted before and
  after: **146 byte-identical, 0 worse**. None of them stores a data page
  uncompressed, so none takes the new path.
- `decode_R2007` is reached only for `R_2007a..R_2007` — r2010+ goes through
  `decode_R2004` — so the blast radius is exactly AC1021 files. **All 47 AC1021
  drawings in a 1657-file corpus**, both builds, same measure:

  | | before | after |
  |---|---:|---:|
  | entities total | 639053 | **667767** (+28714) |
  | drawings gaining | — | 6 |
  | drawings losing | — | **0** |
  | unchanged | — | 41 |

  Three of the six go `LOAD_FAIL -> OK` (+11550, +10847, +5725); the other three
  gain 182–207 entities each.

Built on Ubuntu 26.04, gcc 15.2.0, at `e405fcff` (= tag `0.14.8556`). The
reproducers are already attached to #1356
([handle-collision-dwgs.zip](https://github.com/tuxiasumari/libredwg/releases/download/bugfiles-2026-08-06/handle-collision-dwgs.zip));
`sedapar.dwg` is the 11 MB one, `cofopri.dwg` shows it too.

### DIFF
--- a/src/decode_r2007.c
+++ b/src/decode_r2007.c
@@ -682,6 +682,20 @@ read_system_page (Bit_Chain *out, Bit_Chain *dat, int64_t size_comp,
   return true;
 }
 
+/* Size on disk of a Reed-Solomon coded data page holding size_comp payload
+   bytes: RS works on 0xFB-byte blocks expanded to 0xFF, and the page is then
+   padded to a multiple of 32.  A page whose size matches this exactly is RS
+   coded; a page stored verbatim is only padded to 32 and comes out smaller.
+   The section page record has no flag for it (unlike r2004+), so the size is
+   the only thing to go by.  */
+static int64_t
+page_size_if_rs_coded (const int64_t size_comp)
+{
+  const int64_t pesize = (size_comp + 7) & ~7;
+  const int64_t block_count = (pesize + 0xFB - 1) / 0xFB;
+  return (block_count * 0xFF + 31) & ~31;
+}
+
 static int
 read_data_page (Bit_Chain *restrict dat, BITCODE_RC *restrict decomp,
                 int64_t page_size, int64_t size_comp, int64_t size_uncomp,
@@ -830,9 +844,16 @@ read_data_section (Bit_Chain *sec_dat, Bit_Chain *dat,
           return DWG_ERR_VALUEOUTOFBOUNDS;
         }
       dat->byte = page->offset;
-      // only if compressed. TODO: Isn't there a compressed flag as with 2004+?
-      // theoretically the sizes could still be the same.
-      if (section_page->comp_size != section_page->uncomp_size)
+      /* Reed-Solomon coding is independent of compression: read_data_page()
+         undoes the RS coding first and only then decompresses, when
+         comp_size < uncomp_size.  An uncompressed page can therefore still be
+         RS coded, and copying it straight out of dat->chain hands the caller
+         the RS codewords -- payload interleaved with parity.  Take the RS path
+         whenever the page on disk is exactly as large as the RS coded form of
+         its payload; page->size is the only evidence available. */
+      if (section_page->comp_size != section_page->uncomp_size
+          || (int64_t)page->size
+                 == page_size_if_rs_coded (section_page->comp_size))
         {
           error = read_data_page (dat, &decomp[section_page->offset],
                                   page->size, section_page->comp_size,
@@ -840,7 +861,7 @@ read_data_section (Bit_Chain *sec_dat, Bit_Chain *dat,
           if (error)
             {
               free (decomp);
-              LOG_ERROR ("Failed to read compressed page");
+              LOG_ERROR ("Failed to read page");
               return error;
             }
         }


rurban (approved):
Thanks!

Issue #1012, opened 2024-09-20
problem in dwg2SVG


acabrera1000:
I'm trying dwg2SVG, and I think there is an error in the implementation of 	static int output_BLOCK_HEADER (Dwg_Object_Ref *ref).

When output_object is called, the references of the objects of the block should be to the point of definition of the block, not to
the point of insertion of the block.

rurban:
Sounds right, but I have to check

tuxiasumari:
You are right, and #1365 fixes it.

`transform_X`/`transform_Y` translated every coordinate by the model origin,
including the coordinates inside a `<defs>` symbol — while `output_INSERT`
already places the symbol with
`translate(transform_X(ins_pt), transform_Y(ins_pt))`. So each block was
displaced by the model origin a second time, and by however far the drawing sits
from (0,0).

The definitions are now emitted in their own space (only the y flip, which SVG
needs either way) and the `<use>` alone carries the placement. It also matters
for the framing fix in that PR: only insertion points belong in the box, not
block-local coordinates.

That PR also fixes #523, which had the same visible outcome — a blank page — from
three further causes.

PR #1365, opened 2026-08-07
dwg2SVG: the output was blank for four independent reasons (fixes #523, #1012)

### LINKED ISSUE #523: dwg2SVG wrong viewBox
Hi, I am using dwg2SVG from docker. Have not managed to use this command for any of the files in the test folder [libredwg](https://github.com/LibreDWG/libredwg)/[test](https://github.com/LibreDWG/libredwg/tree/master/test)/test-data/. For example:
dwg2SVG example_r14.dwg > example_r14.svg this gives an empty content file as attached.
example_r14
or dwg2SVG example_2018.dwg > example_2018.svg
example_2018
.dwg files and .svg output , just in case
[examples 2.zip](https://github.com/LibreDWG/libredwg/files/9989957/examples.2.zip)

tuxiasumari:
Fixes #523 and #1012. Reproducer: `dwg2SVG test/test-data/example_r14.dwg`, which
is where the report starts — LibreDWG's own test file.

The page was blank for **four independent reasons**. Each one alone is enough.

## 1. Model space was never drawn

`output_BLOCK_HEADER` returns *"how many paper-space entities we did print"*, and
`output_SVG` falls back to model space when that is zero:

```c
  if (!mspace && (ref = dwg_paper_space_ref (dwg)))
    num = output_BLOCK_HEADER (ref);
  if (!num && (ref = dwg_model_space_ref (dwg)))
    output_BLOCK_HEADER (ref);
```

But `output_object` starts at `int num = 1` and only the `default` branch clears
it — so it returns 1 for **every** type it recognises, including `VIEWPORT` and
`SEQEND`, which draw nothing, and entities it skips as invisible. Every paper
space holds a `VIEWPORT`, so the count was never zero and the fallback never
fired.

The model-space geometry *was* in the file — but only inside `<defs>`, because
`*Model_Space` is also a `block_control` entry. **No renderer draws the contents
of `<defs>`.** That is why it looked empty while the file was full of paths, and
why @insertintosb found in this thread that deleting the `<defs>` tags made the
drawing appear.

The walk now counts entities that actually put coordinates on the page.

## 2. Every stroke was zero width

```c
  int lw = dxf_cvt_lweight (ent->linewt);
  return lw < 0 ? 0.1 : (double)(lw * 0.001);
  ...
  printf ("...stroke-width:%.1fpx...", lweight);
```

`dxf_cvt_lweight` gives 100ths of a millimetre, so millimetres are `lw/100`, not
`lw/1000`. With the extra factor of ten a 0.25 mm line became 0.025, which
`%.1f` prints as **`stroke-width:0.0`**. And `lw == 0` is a legitimate value
meaning "thinnest line", so it needs a width of its own too.

Widths are now emitted in mm with `vector-effect:non-scaling-stroke`, which is
what keeps a hairline visible whatever the viewBox spans.

## 3. The viewBox was the header's box, not the drawing's

`$EXTMIN`/`$EXTMAX` cover the whole drawing, including entity types this program
skips, and real files carry entities outside their recorded extents as well. For
`example_r14.dwg` the header box is **3540706 × 2726367** while the geometry
actually emitted spans about **14871 × 9702** — everything collapsed well below
a pixel. @rurban named the symptom here in 2022:

> `viewBox="-2680430.068109 -1672547.179548 3540706.089899 2726367.694617"` is
> way too large, and you cannot see anything there

The drawn space is now walked twice: once with output going to the null device to
collect the box, then again to emit. It cannot simply *skip* the `printf` calls
during the first pass — the arguments are where `transform_X`/`transform_Y` live,
so not evaluating them measures nothing. (I made that mistake first; the
measured box came back as `DBL_MAX`.)

## 4. The viewBox origin contradicted the coordinates

`transform_X` returns `x - model_xmin` and `transform_Y` returns
`page_height - (y - model_ymin)`, so every emitted coordinate lives in
`[0,page_width] × [0,page_height]`. The viewBox started at
`(model_xmin, model_ymin)` instead of `(0,0)`, displacing the content out of view
by exactly the model origin. Invisible in a drawing whose `$EXTMIN` is `(0,0)`,
fatal for the UTM survey drawings where it is not.

## And #1012

@acabrera1000 is right. Inside a `<defs>` symbol the coordinates are
block-local, and the `<use>` that instantiates it already carries
`translate(transform_X(ins_pt), transform_Y(ins_pt))` — so translating the
definition as well displaced every block by the model origin a **second** time.
Definitions are now left in their own space, which is also what makes the
measured box meaningful (only insertion points belong in it).

## Result

`example_r14.dwg` renders: border, pentagon, ellipse, arc, circles, hatching.

On **95 real drawings** from a 1657-file corpus, each rendered to a 200×200 PNG
and scored by how many pixels carry ink:

| | stock | with this patch |
|---|---:|---:|
| drawings that show a drawing | **1** | **68** |
| of those, blank before | — | 67 |
| **drawings that got worse** | — | **0** |
| ink pixels, total | 80 | 55775 |

## Verification

- `make check`: **270 PASS / 0 FAIL / 0 SKIP**, unchanged — and the dwg2SVG
  tests are in it.
- All **148** DWGs in `test/test-data` and `programs/` still convert with a zero
  exit status and still parse as well-formed XML.

## Two things I did not touch

- `output_XLINE` prints its ray parameters (`txmin, tymin, txmax, tymax`)
  instead of the clipped endpoints, so an XLINE comes out as a path with
  meaningless coordinates. The function is marked `// untested!` and its clip box
  is built with `model_ymin` twice (`box[0].y` and `box[1].y`). Separate fix, and
  I would rather not guess at the intended geometry.
- Drawings where one INSERT sits 25 million units from the rest still come out as
  a speck: framing the union of everything drawn is then dominated by the
  outlier. Bounding that is a policy decision about outliers that upstream should
  own — say the word if you want a percentile-based box and I will measure one.

Happy to split this into four commits, or into separate PRs, if that is easier to
review. Built on Ubuntu 26.04, gcc 15.2.0, at `e405fcff` (= tag `0.14.8556`).

### DIFF
--- a/programs/dwg2SVG.c
+++ b/programs/dwg2SVG.c
@@ -47,6 +47,7 @@
 #endif
 #include "my_getopt.h"
 #include <math.h>
+#include <float.h>
 #ifdef HAVE_VALGRIND_VALGRIND_H
 #  include <valgrind/valgrind.h>
 #endif
@@ -105,6 +106,55 @@ help (void)
   return 0;
 }
 
+/* Every geometry printf goes through this. The measuring pass below has to
+   walk the same code without emitting, and it cannot simply skip the call:
+   the arguments are where transform_X and transform_Y live, so not evaluating
+   them measures nothing. So it writes to the null device instead.
+   fprintf (stderr, ...) is untouched. */
+#ifdef _WIN32
+#  define NULL_DEVICE "NUL"
+#else
+#  define NULL_DEVICE "/dev/null"
+#endif
+static FILE *measure_sink = NULL;
+
+#define printf(...) fprintf (measuring ? measure_sink : stdout, __VA_ARGS__)
+
+/* The viewBox has to be the box of what actually gets drawn, and that is not
+   $EXTMIN/$EXTMAX: this program renders "some 2D elements", so the header box
+   also covers entities it skips, and real drawings carry entities outside
+   their recorded extents too. Framing by the header box left the geometry off
+   screen or smaller than a pixel -- the empty SVGs of GH #523.
+
+   So the drawn space is walked twice: once with output suppressed to collect
+   the box, then again to emit. Coordinates are recorded before the transform,
+   since the transform itself needs the box.
+
+   Inside a <defs> symbol the coordinates are block-local -- the <use> that
+   instantiates it already carries the placement -- so translating them there
+   as well displaced every block by the model origin a second time. They are
+   left alone on the way out, and kept out of the box: what belongs in it is
+   the insertion point, which the <use> does pass through transform_X. */
+static bool measuring = false;
+static bool in_block_definition = false;
+static double drawn_xmin, drawn_ymin, drawn_xmax, drawn_ymax;
+
+/* Counts coordinates put on the page, so output_BLOCK_HEADER can tell which
+   of its entities actually drew something. */
+static unsigned long coords_drawn = 0;
+
+static void
+record_XY (double v, double *lo, double *hi)
+{
+  if (!isfinite (v) || in_block_definition)
+    return;
+  coords_drawn++;
+  if (v < *lo)
+    *lo = v;
+  if (v > *hi)
+    *hi = v;
+}
+
 static double
 transform_ANGLE (double angle)
 {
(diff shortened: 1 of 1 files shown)


rurban (approved):
Great. Just waiting for the CLA

PR #1387, opened 2026-08-09
out_dxf: preserve an unparseable entity as a proxy instead of dropping it

tuxiasumari:
### Symptom

An entity whose class LibreDWG cannot parse (`UNKNOWN_ENT` — Civil 3D road
objects, custom-application entities) is returned as `DWG_ERR_UNHANDLEDCLASS`
by `dwg_dxf_object` and **vanishes from the DXF entirely**, taking its proxy
graphics with it. ODA-based converters instead preserve it as an
`ACAD_PROXY_ENTITY` carrying the graphics blob, so viewers draw the geometry
even without the object enabler.

### Fix

Emit the same when the class is unparseable: the common entity data, the proxy
class markers (`90`=498, `91`=class id, `95`=version, `70`=0), and the proxy
graphics in the proxy-data group (`92` size + `310` hex).

**Round-trip safe.** LibreDWG's own `in_dxf` reads that back via
`add_ent_preview`, so the DXF still round-trips through `dxf2dwg` — verified by
`make check` (`dxf.test` re-imports `example_2000`, which carries a proxy). One
subtlety worth noting for reviewers: group 95 must be emitted **decimal**, not
hex — `dxf_read_rl` parses it decimal, and a hex value desyncs the reader.

The class data section (group 93 stream + its object-id reference lists) is
**not** emitted: it is not reconstructible without parsing the class, and is
useless to a reader without the object enabler anyway. `data_numbits` is
written 0 to keep the record well-formed; the graphics blob is what viewers
need.

### Measured

Over 1657 real DWGs (`dwg2dxf` → ezdxf recover):
- **181 files gain entities, +112090 total** — proxy geometry newly preserved;
- **0 files lose any entity** — no regression;
- two Civil 3D drawings that were empty salvage now convert.

`make check` green (254 tests); compiles standalone on a clean master tree.
Stacks naturally on the DXF-emission-robustness fixes (#1373 etc.) which let
emission reach the entities section of such drawings in the first place.

### DIFF
--- a/src/out_dxf.c
+++ b/src/out_dxf.c
@@ -2792,6 +2792,60 @@ dxf_3dsolid (Bit_Chain *restrict dat, const Dwg_Object *restrict obj,
 
 /* returns 0 on success
  */
+/* An entity of a class we cannot parse still carries its common entity
+   data and, usually, its proxy graphics blob. Emit it as ACAD_PROXY_ENTITY
+   with that blob — as ODA-based converters do — so conformant viewers draw
+   the graphics instead of silently losing the entity. The class data bits
+   are passed through in DWG form (group 93/310), like our PROXY_ENTITY
+   writer does. */
+static int
+dwg_dxf_UNKNOWN_ENT_as_proxy (Bit_Chain *restrict dat,
+                              const Dwg_Object *restrict obj)
+{
+  int error = 0;
+  const Dwg_Data *dwg = obj->parent;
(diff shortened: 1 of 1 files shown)


rurban:
That is a very good idea

rurban (approved):

tuxiasumari:
Gentle ping: this one is approved but still open. Happy to rebase it onto current master if that helps — it applies cleanly as of 2d8f277.

PR #1388, opened 2026-08-10
Improve repeat and endrep work

michal-josef-spacek:


michal-josef-spacek:
DONE: Need to fix mingw build

PR #1401, opened 2026-08-21
bound the codepage index in codepage_helper and dwg_codepage_isalnum

saddamr3e:
UBSan on a DWG with an out-of-range header.codepage:

    src/codepages.c:312:11: runtime error: index 5000 out of bounds for type 'const uint16_t *[46]'
    src/codepages.c:445:32: runtime error: index 5000 out of bounds for type 'const uint8_t *[46]'

header.codepage is read verbatim (header.spec / 2ndheader.spec) and used as the index into the 46-entry cp_fntbl / cp_alnumtbl / cp_alnum16tbl tables (valid 0..CP_ANSI_1258). codepage_helper and the default branch of dwg_codepage_isalnum index and dereference without a range check, so an out-of-range value reads past the array and derefs a wild pointer. It reaches the helpers from the pre-r2007 TV->UTF-8 conversion (bit_TV_to_utf8_codepage, the non-iconv path) and the name validators in dwg_api.c. Reject the out-of-range codepage in both callees.

Regression test added to bits_test (codepage_bounds_tests): fails under the UBSan CI job before the fix, passes after.

### DIFF
--- a/src/codepages.c
+++ b/src/codepages.c
@@ -309,6 +309,8 @@ codepage_helper (const Dwg_Codepage codepage, const wchar_t wc, const int dir,
   uint16_t maxc;
   assert (codepage != CP_UTF8 && codepage != CP_UTF16
           && codepage != CP_US_ASCII && codepage != CP_ISO_8859_1);
+  if ((unsigned)codepage > CP_ANSI_1258 || !cp_fntbl[codepage])
+    return 0;
   fntbl = cp_fntbl[codepage];
   maxc = fntbl[0];
   assert (maxc);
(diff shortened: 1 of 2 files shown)


rurban (approved):
Thanks, but the tests exceed 15 lines. CLA needed

saddamr3e:
Trimmed the test down to a single 12-line check (still covers codepage_helper via dwg_codepage_uc/dwg_codepage_wc plus dwg_codepage_isalnum), so the addition is under the 15-line threshold now. If you'd still prefer the FSF assignment for future patches, point me at the form and I'll start the paperwork.

rurban:
The CLA URL is in CONTRIBUTING

PR #1402, opened 2026-08-21
Three quadratic costs in large imports and dwg_add sessions (fixes #1395)

### LINKED ISSUE #1395: Performance: dwg_insert_entity can become quadratic for large blocks
## Problem

Adding many entities to the same `BLOCK_HEADER` through the public `dwg_add_*` API becomes very slow as the number of entities grows.

The slowdown happens while LibreDWG builds the block entity links in memory, before the DWG file is written to disk.

## Root cause

Every `dwg_add_*` call invokes `dwg_insert_entity()`.

For every top-level entity, `dwg_insert_entity()` appends its handle to `BLOCK_HEADER.entities` using `PUSH_HV`.

`PUSH_HV` reallocates the array to exactly `num_owned + 1` elements on every append:

```c
_obj->hvfield = (BITCODE_H *)realloc(
    _obj->hvfield,
    (_obj->numfield + 1) * sizeof(BITCODE_H));
```

There is no separately tracked allocation capacity.

When `realloc()` cannot extend the existing allocation in place, it must move and copy the complete array. Entity creation performs other allocations between appends, so this relocation happens frequently in large drawings.

In the worst case, adding `N` entities requires copying approximately:

```text
1 + 2 + 3 + ... + N
```

handles, resulting in O(N²) behavior.

Relevant source code:

- https://github.com/LibreDWG/libredwg/blob/master/src/dwg_api.c#L24659-L24698
- https://github.com/LibreDWG/libredwg/blob/master/src/common.h#L595-L606

## Measurements

The following results were measured with the same downstream writer and data. The only relevant difference was how the `BLOCK_HEADER` entity links were constructed:

| Number of objects | Normal linking | Deferred and batched linking |
|---:|---:|---:|
| 136,794 | 19.19 seconds | 3.18 seconds |
| 820,953 | 319.8 seconds | 53.7 seconds |

## Proof of concept

As a proof of concept, automatic link maintenance was temporarily deferred for top-level `BLOCK_HEADER` entities.

The entity handles were collected in a pre-reserved vector. After all entities had been created, the following data was constructed in one pass:

- `BLOCK_HEADER.entities`
- `BLOCK_HEADER.num_owned`
- `BLOCK_HEADER.first_entity`
- `BLOCK_HEADER.last_entity`
- Entity `prev_entity` and `next_entity` links

This removes the repeated growth of `BLOCK_HEADER.entities`.

The proof of concept currently uses a private option flag, but I do not propose that project-specific flag as the final upstream API.

## Proposed upstream solution

Possible general solutions are:

1. Add a batch insertion API with an expected entity count.
2. Add capacity-aware growth for `BLOCK_HEADER.entities`.
3. Provide another supported way to defer and finalize block entity links.

tuxiasumari:
Fixes #1395, whose analysis of `PUSH_HV` is confirmed — and profiling the reproducer turned up two more independent quadratic costs on the same path. One commit per root cause, 33 lines total:

1. **`PUSH_HV` / `PUSH_HV_NEW` realloced to exactly num+1 on every push** (the cost #1395 diagnosed). A handle vector has no tracked capacity and some call sites allocate one to its exact size, so the realloc-per-push stays — but the requested size is now rounded up to the next power of two: calls between two crossings resolve to the same usable size and do not copy, and realloc knows the vector's real old size, so the push is correct however the vector was last allocated. N pushes copy O(N) handles instead of O(N²). This dominates on allocators without mremap-style in-place growth (it is why the issue's measurements are much worse than what glibc shows).

2. **`Ref_cmp` compared `handleref.code` first**, so a new ref — whose handle is the highest so far — sorted to the end of its code block, in the *middle* of the ordered index, and every `ordered_ref_add` memmoved all higher-code refs one slot up. With `absolute_ref` as the primary key a new ref lands at the end and the memmove copies nothing. The order is private to `ordered_ref_add`/`ordered_ref_find`, which both use this comparator.

3. **`dwg_decode_add_object_ref` set `dirty_refs = 1` every time the ref pointer array grew by a chunk** (16384 refs). That realloc moves only the array of *pointers*; the refs and their cached `ref->obj` are untouched. Its sibling `dwg_add_object` already demonstrates the correct criterion — dirty only when `dwg->object` itself moved. The spurious flag made the next `dwg_ref_object` run a full `dwg_resolve_objectrefs_silent` sweep, once per 16384 refs: gprof shows 36 full sweeps / 7.6M resolves for 120k adds. This was the dominant cost on glibc.

**Measured** (loop of `dwg_add_LINE` into modelspace, glibc/Linux, -O2):

| N | master | patched |
|---:|---:|---:|
| 40 000 | 0.16 s | 0.07 s |
| 160 000 | 2.09 s | 0.39 s |
| 640 000 | 29.62 s | 1.84 s |

Per-add cost is now flat (µs/add 1.7 → 2.9 over 16× N; master grows 3.9 → 46.3). `dxf2dwg --as r2000` of a 300k-entity DXF: 2.49 s → 1.71 s, output byte-identical.

**Verified**: `make check` all green; a 1657-file real-DWG corpus converts with identical entity counts and content signatures to master (0 regressions); a 400-seed DXF→DWG→DXF round-trip fuzz campaign is result-identical to master, seed by seed; the DWGs written from large DXFs are byte-identical.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -482,6 +482,24 @@ const unsigned char _dwg_VISUALSTYLE_proptypes[58] = {
   /* [56] */ 3, 3
 };
 
+/* A handle vector has no separately tracked capacity, and some call sites
+   (re)allocate one to its exact size.  So PUSH_HV keeps its realloc on every
+   push, but rounds the requested size up to the next power of two: realloc
+   knows the vector's real old size, so the push stays correct however the
+   vector was last allocated, and the calls between two crossings resolve to
+   the same size and do not copy.  N pushes then copy O(N) handles instead of
+   the O(N^2) an exact num+1 realloc costs (GH #1395).
+   Deliberately not static inline in common.h: on a cold path gcc declines to
+   inline it, and -Werror=inline then fails the build (mips32 CI).  */
+size_t
+hv_alloc_size (const size_t num)
+{
+  size_t cap = 8;
+  while (cap < num)
+    cap <<= 1;
+  return cap * sizeof (BITCODE_H);
+}
+
 // returns the first ref from the handle vector.
 BITCODE_H
 shift_hv (BITCODE_H *hv, BITCODE_BL *num_p)
--- a/src/common.h
+++ b/src/common.h
@@ -630,12 +630,16 @@ void *memmem (const void *h0, size_t k, const void *n0, size_t l)
 #  endif
 #endif
 
+/* The byte size to (re)allocate a handle vector of num entries with.
+   See src/common.c.  */
+size_t hv_alloc_size (size_t num);
+
 // push to handle vector at the end. It really is unshift.
 #define PUSH_HV(_obj, numfield, hvfield, ref)                                 \
   if (_obj->numfield <= 0 || _obj->hvfield[_obj->numfield - 1] != ref)        \
     {                                                                         \
       _obj->hvfield = (BITCODE_H *)realloc (                                  \
-          _obj->hvfield, (_obj->numfield + 1) * sizeof (BITCODE_H));          \
+          _obj->hvfield, hv_alloc_size (_obj->numfield + 1));                 \
       _obj->hvfield[_obj->numfield] = ref;                                    \
       LOG_TRACE ("%s[%d] = " FORMAT_REF " [H]\n", #hvfield, _obj->numfield,   \
                  ARGS_REF (_obj->hvfield[_obj->numfield]));                   \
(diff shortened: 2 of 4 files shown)


tuxiasumari:
Force-pushed: the mips32-big-endian build was red on my account, and I fixed it.

`hv_alloc_size` was a `static inline` in `common.h`, but `PUSH_HV` sits on cold paths, so gcc declined to inline it there and `-Werror=inline` failed that target. It is now defined out of line in `src/common.c`, next to `shift_hv`/`delete_hv`/`find_hv`, with the prototype in `common.h` — the same pattern those three already use. The function body and both call sites are unchanged; the net diff against the previous head is only the move.

Since it carries no `inline` keyword any more, `-Winline` cannot fire on it by construction. Verified locally with `-O2 -Winline -Werror=inline` on every translation unit that uses `PUSH_HV`/`PUSH_HV_NEW` (`common.c`, `in_dxf.c`, `decode.c`, `decode_r11.c`, `dwg.c`, `dwg_api.c`): all clean, and `common.c` is also clean under `-Wall -Wextra -Wmissing-prototypes -Wredundant-decls -Wstrict-prototypes -Werror`.

Squashed into the commit that introduced the helper rather than added on top, as there was no review to orphan.

rurban (approved):

PR #1407, opened 2026-08-29
decode: read MTEXT background scale as BD

encku:
## Summary

- decode the R2018 MTEXT background-fill scale (DXF group 45) as `BD`
  instead of `BL`
- keep the public struct, ODA reference patch, unit test, and generated
  dynapi files consistent with that type
- preserve bitstream alignment for the following background-color and
  annotative-data fields

## Root cause

MTEXT background-fill scale is fractional and defaults to `1.5`. Reading it
as a bit-long consumes the wrong bit pattern. The remaining MTEXT fields are
then decoded from the wrong positions, which can produce invalid color data,
an invalid class version, and corrupt later values.

## Public reproducer

This creates a synthetic file; no private drawing is required:

```python
import ezdxf

doc = ezdxf.new("R2018")
mtext = doc.modelspace().add_mtext(
    "Synthetic MTEXT",
    dxfattribs={"insert": (12.5, 34.5, 0.0), "char_height": 2.5},
)
mtext.set_bg_color(7, scale=1.5)
doc.saveas("synthetic_mtext.dxf")
```

I converted this DXF to an R2018 DWG with ODA File Converter 27.1:

```sh
ODAFileConverter input output ACAD2018 DWG 0 1 synthetic_mtext.dxf
```

Converting that DWG back to DXF with `dwg2dxf` gives:

| revision | group 45 | insertion point | decoder errors |
| --- | ---: | --- | --- |
| `7f46cfa4` | `0.0` | `(12.5, 34.5, 0.0)` | two invalid `BL` codes, invalid CMC flag/method, invalid MTEXT class version |
| this branch | `1.5` | `(12.5, 34.5, 0.0)` | none |

The resulting DXF was read with ezdxf 1.4.3 for both measurements.

## Private corpus validation

I also tested four non-distributable R2018 inputs. The counts of invalid DXF
tags reported after conversion changed from `26`, `22`, `28`, and `50` to
zero for all four inputs. No private input, output, filename, path, handle,
coordinate, or hash is included in this PR.

## Tests

- `make check` (passes; 254 unit tests passed)
- `INPUT=synthetic_mtext.dwg test/unit-testing/mtext`
- `INPUT=synthetic_mtext.dwg test/unit-testing/dynapi_test`
- synthetic ODA 27.1 round trip checked with ezdxf 1.4.3


rurban (approved):

rurban:
Thanks! Helps a lot

